PR #182, opened 2015-03-03
Fix missing or stray double quotes in design-data.

rsto:
The properties and value types templates in design-data seem to contain erroneously quoted strings.

I am new to libical, so sorry if these quotes are set intentionally as-is. Using the proposed pull request, the generated C code in icalderivedproperty.c does not differ from the original, so the current mkderivedproperties.pl script already seems robust enough to handle stray and missing quotes. Still, fixing these odd looking quotes now might prevent trouble in future changes of the CSV parser.


winterz:
thanks

PR #188, opened 2015-04-10
Fixed an infinite loop in icalrecur_iterator

ksuther:
Iterating over a yearly recurrence rule that generates no instances was getting stuck in an infinite loop.

For example, this rule generates no occurrences since there's no February 31: FREQ=YEARLY;COUNT=2;BYMONTH=2

icalrecur_iterator_new would give up when hitting the year 20000, but icalrecur_iterator_next wouldn't recognize that and return icaltime_null_time.

Adding a check for icalrecur_iterator_new's behavior seems to be the simplest fix for this.


ksuther:
I don't use libical with ICU right now, so I don't know the implications of making changes when compiling with that enabled. I don't see icalrecur_iterator_new calling down into anything related to ICU (at least based on the ifdefs), so would the patch you proposed work for it?

The context that I saw this recurrence rule was with a DTSTART of 19410131. Other calendars seem to handle this by showing just the first instance with no recurrences (show 19410131, no future years have anything in February).

ksmurchison:
The first patch that I proposed won't work when compiled with ICU but I won't bore you with the reasons.  What was the actual example with a DTSTART 0f 19410131?  That is a valid date and as far as I know should work with any RRULE.

Have you actually seen a DTSTART of 19410231 in the wild?

ksuther:
Yeah, the actual recurrence and start date I encountered in the wild is 19410131 (what I put into the regression). Unfortunately I don't have the entire ics blob for it, as it came from a user and I had minimal logging on to track down just this problem.

ksmurchison:
I'm confused. Are you saying that 19410131 caused a problem?  What was the RRULE?

I don't have a problem with DTSTART that doesn't match the RRULE pattern. My thought is that we should reject nonsensical DTSTART like 19410231

ksuther:
Sorry, I'll try that again.

The RRULE was FREQ=YEARLY;COUNT=2;BYMONTH=2
DTSTART was 19410131

It was then trying to expand into 19420231, etc (subsequent years with February 31). So the DTSTART started out as sensible, but when trying to apply it to the RRULE it resulted in no valid occurrence dates.

ksmurchison:
Ok. Let me test that tomorrow and see where it was failing

ksmurchison:
OK.  I verified that my patch fixes the problen with the RRULE above and if DTSTART is a bogus value (e.g 20150229, 20150132, 20151301) both with and without ICU.  I have committed these changes to master.

winterz:
awesome work guys. thanks!

ksuther:
Thanks for the improvements Ken!

PR #203, opened 2015-06-16
Use bsearch in icalparameter_string_to_kind

ksuther:
Use bsearch in icalparameter_string_to_kind instead of linearly searching all parameters. This matters more here since it calls strcasecmp instead of doing an equality check like most of the other string_to_kind or kind_to_string functions.

mkderivedparameters.pl explicitly sorts parameters by their name, so it's safe to use binary search here.


ksmurchison:
This change makes sense to me.

winterz:
I'll merge this. but I already know the compile will fail with MS Visual Compilers. I'll fix that after the merge.

PR #204, opened 2015-06-17
portability: define bswap_32 if not defined already

ajacoutot:
Also define bswap_32 in the HAVE_ENDIAN_H case; this is needed at least on
OpenBSD to prevent an undefined reference during build.


ajacoutot:
You're right, sorry I missed it.
I just created a new pull request that should fix that.

PR #205, opened 2015-06-18
On Windows, the bswap_32 macro is defined later on.

ajacoutot:
That's a follow up of https://github.com/libical/libical/pull/204 .
Sorry for missing it.


winterz:
I notice that FreeBSD has the bswapXX functions defined in machine/bswap.h

does OpenBSD also have machine/bswap.h?

ajacoutot:
Hi Allen.
No, OpenBSD does not bundle such header. We have no definition for bswapXX, only swapXX.

winterz:
then how about if we go back to what we had previously, but add 
#if defined(__OPENBSD__)
#define bswap_32 swap32
#endif
before the #if defined(_MSC_VER) section

ajacoutot:
Oh I am absolutely fine with me yes. I originally tried to make it a bit more generic, but your proposal will work just fine :-)
You need to use this though:
#if defined(__OpenBSD__)

PR #210, opened 2015-08-10
icalparser optimizations

ksuther:
The following changes result in approximately a 15-20% speedup in icalcomponent_new_from_string when loading a large ics file with lots of parameters on each property (> 1 MB).

parser_get_next_char was performing a bunch of redundant pointer loads that clang wasn't able to optimize out. Reducing those made it possible for clang to eliminate the function completely (since it is static) when optimizations are enabled.

parser_get_param_name was always malloc/freeing strings inside icalparser_add_line, which added up over time. I added a second version that uses stack-allocated strings when possible, which reduced the number or redundant allocations. parser_get_prop_name and parser_get_next_value could use the same treatment as they're now showing up as heavier functions.

### DIFF
--- a/src/libical/icalcomponent.c
+++ b/src/libical/icalcomponent.c
@@ -97,16 +97,13 @@ static icalcomponent *icalcomponent_new_impl(icalcomponent_kind kind)
         return 0;
     }
 
+    memset(comp, 0, sizeof(icalcomponent));
+
     strcpy(comp->id, "comp");
 
     comp->kind = kind;
     comp->properties = pvl_newlist();
-    comp->property_iterator = 0;
     comp->components = pvl_newlist();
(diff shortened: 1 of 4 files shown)


winterz:
you need to fix the builds with gcc that are failing on strlcpy()

see https://travis-ci.org/libical/libical/jobs/74999133

winterz:
http://stackoverflow.com/questions/18547251/when-i-use-strlcpy-function-in-c-the-compilor-give-me-an-error discusses strlcpy(), strncpy(), memcpy() , etc.

ksuther:
Whoops, let's see how Linux likes strncpy.

winterz:
I need to test on Windows still

PR #216, opened 2015-09-03
Set icalerrno on out-of-bounds write in icalrecur_iterator_impl->days[]

champtar:
~~This is not a real fix (we get a SIGABRT),~~
~~but this prevent memory corruption~~

Spotted thanks to AFL (American Fuzzy Lop)

Signed-off-by: Etienne CHAMPETIER <champetier.etienne@gmail.com>

### DIFF
--- a/src/libical/icalrecur.c
+++ b/src/libical/icalrecur.c
@@ -2820,8 +2820,13 @@ static int expand_year_days(icalrecur_iterator *impl, int year)
             /* FREQ=YEARLY; */
             int doy = get_day_of_year(impl, year, 0, 0, NULL);
 
-            if (doy != 0)
+            if (doy != 0) {
+                if (days_index >= ICAL_BY_YEARDAY_SIZE) {
+                    icalerror_set_errno(ICAL_INTERNAL_ERROR);
+                    return 0;
+                }
                 impl->days[days_index++] = (short)doy;
+            }
 
             break;
         }
@@ -2834,8 +2839,13 @@ static int expand_year_days(icalrecur_iterator *impl, int year)
 
                 doy = get_day_of_year(impl, year, month, 0, NULL);
 
-                if (doy != 0)
+                if (doy != 0) {
+                    if (days_index >= ICAL_BY_YEARDAY_SIZE) {
+                        icalerror_set_errno(ICAL_INTERNAL_ERROR);
+                        return 0;
+                    }
                     impl->days[days_index++] = (short)doy;
+                }
             }
             break;
         }
@@ -2848,8 +2858,13 @@ static int expand_year_days(icalrecur_iterator *impl, int year)
 
                 doy = get_day_of_year(impl, year, 0, month_day, NULL);
 
-                if (doy != 0)
+                if (doy != 0) {
+                    if (days_index >= ICAL_BY_YEARDAY_SIZE) {
+                        icalerror_set_errno(ICAL_INTERNAL_ERROR);
+                        return 0;
+                    }
                     impl->days[days_index++] = (short)doy;
+                }
             }
             break;
         }
@@ -2865,8 +2880,13 @@ static int expand_year_days(icalrecur_iterator *impl, int year)
 
                     doy = get_day_of_year(impl, year, month, month_day, NULL);
 
-                    if (doy != 0)
+                    if (doy != 0) {
+                        if (days_index >= ICAL_BY_YEARDAY_SIZE) {
+                            icalerror_set_errno(ICAL_INTERNAL_ERROR);
+                            return 0;
+                        }
                         impl->days[days_index++] = (short)doy;
+                    }
                 }
             }
 
@@ -2906,6 +2926,11 @@ static int expand_year_days(icalrecur_iterator *impl, int year)
             for (i = pvl_head(days); i != 0; i = pvl_next(i)) {
                 short day = (short)(intptr_t) pvl_data(i);
 
+                if (days_index >= ICAL_BY_YEARDAY_SIZE) {
+                    icalerror_set_errno(ICAL_INTERNAL_ERROR);
+                    pvl_free(days);
+                    return 0;
+                }
                 impl->days[days_index++] = day;
             }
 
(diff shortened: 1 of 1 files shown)


echampet:
hi @winterz,

can you take a look at this PR
Unit tests are failing with or without my patch on my computer,
and it's only asserts (so if we do hit them we have to fix the code, not the assert :) )

champtar:
hi @winterz,

can you take a look at this PR
Unit tests are failing with or without my patch on my computer,
and it's only asserts (so if we do hit them we have to fix the code, not the assert :) )

ksmurchison:
Is this a preemptive patch, or was code actually failing?  If so, where?

champtar:
hi @ksmurchison,

i'll push another version soon,
using icalerror_set_errno instead of assert so we can recover

I'll also include some exemples to crash libical

champtar:
You can also take a look at my afl branch
https://github.com/champtar/libical/tree/afl
it contains my inputs (all my caldav rrule after afl-cmin) in afl/in/
and a small afl wrapper (i'm mostly interested in icalrecur*) in afl/icalrecur_iterator.c
(command to build at the top)
I can explain more if you are interrested

Please review/merge

ksmurchison:
I'm still not seeing how days_index gets out of range.  Which one of your tests forces it out of range?

champtar:
@ksmurchison,

files in ```afl/in``` are good/normal inputs
please look at my commit message for a crashing exemple

ksmurchison:
I tried the example in your commit message and it didn't crash.  completed with no output

champtar:
without my patch ?

ksmurchison:
Yes

ksmurchison:
Actually, the example in your commit should fail to parse since 0 isn't a valid day.  But I don't think the parser kicks out on bad BYDAY input.

ksmurchison:
The only way that I can see that days_index can get out of bounds is if we allow any of the BY* arrays to exceed their max size and the parser seems to enforce the correct size in icalrecur_add_byrules()

champtar:
I'm getting a sigbus


i'm building with:
```
cmake . -DSTATIC_ONLY=true -DUSE_BUILTIN_TZDATA=true -DWITH_CXX_BINDINGS=false -DCMAKE_BUILD_TYPE=Debug
make
```

can you try (after applying my patch)
```
#include <stdio.h>
#include <stdlib.h>
#include <string.h>
#include "ical.h"

int main()
{
    struct icalrecurrencetype recur = icalrecurrencetype_from_string("FREQ=YEARLY;BYDAY=,,,,,,,,0,,,TH");
    if (icalerrno != ICAL_NO_ERROR) {
        printf("1\n");
        exit(1);
    }
    struct icaltimetype start = icaltime_from_string("20130326T000000");
    if (icalerrno != ICAL_NO_ERROR) {
        printf("2\n");
        exit(2);
    }
    icalrecur_iterator_new(recur,start);
    if (icalerrno != ICAL_NO_ERROR) {
        printf("3\n");
        exit(3);
    }

    exit(0);

}
```
you will see that you hit 3

ksmurchison:
Can you tell me exactly where it crashes?  Do you have a core file that you can backtrace?

champtar:
In any case, we should not write outside days[]
my assert() version (so even if it doesn't crash for you, it will assert): https://github.com/champtar/libical/tree/assert

the backtrace with gdb:
```
#0  0x0000000000420931 in pvl_remove (L=0x665010, E=0x665fa0) at /home/etienne/libical/src/libical/pvl.c:336
#1  0x0000000000420b6a in pvl_clear (l=0x665010) at /home/etienne/libical/src/libical/pvl.c:442
#2  0x000000000042042e in pvl_free (l=0x665010) at /home/etienne/libical/src/libical/pvl.c:84
#3  0x00000000004164ef in expand_year_days (impl=0x665060, year=2013) at /home/etienne/libical/src/libical/icalrecur.c:2912
#4  0x000000000041488b in icalrecur_iterator_new (rule=..., dtstart=...) at /home/etienne/libical/src/libical/icalrecur.c:2104
#5  0x00000000004114f5 in main () at /home/etienne/libical/afl/icalrecur_iterator.c:22
```
it's not obvious that a segfault in pvl_remove come from out of bound write in days[], but it's the cause :)
with my patch i can't crash libical anymore (with https://github.com/champtar/libical/blob/afl/afl/icalrecur_iterator.c and afl-fuzz)

btw, i'm using an ubuntu 14.04 lxc on a centos 6.7 host

ksmurchison:
Which case in the switch statement includes line 2912 in your copy of icalrecur.c?

champtar:
https://github.com/champtar/libical/blob/8da80f6ecdd7d0017a7b16530f48fe054906988f/src/libical/icalrecur.c#L2912

champtar:
(i'm using current master, 8da80f6ecdd7d0017a7b16530f48fe054906988f)

champtar:
can you try to checkout
```
https://github.com/champtar/libical/tree/crash
```
and do:
```
git clean -fd
cmake . -DSTATIC_ONLY=true -DUSE_BUILTIN_TZDATA=true -DWITH_CXX_BINDINGS=false -DCMAKE_BUILD_TYPE=Debug
make
bin/crash
```

on my fedora22 it failed to build because cmake found the system libical,
but in a fedora22 lxc on my fedora22, it build and crash

Are you sure your environment is clean (no libical)?

ksmurchison:
The problem is actually twofold:
First, we are accepting bogus BYDAY data which is getting translated into ICAL_NO_WEEKDAY values in the byday array.  We need to fail to parse bogus data
Second, because we are allowing multiple ICAL_NO_WEEKDAY values and we don't check for and eliminate duplicate days in expand_by_day() we exceed ICAL_BY_YEARDAY_SIZE

I will work to fix both of these problems.

champtar:
ping me when you are done,
i'll launch my fuzzer again to see if it's all fixed

ksmurchison:
I just committed my fix for preventing the out-of-bounds problem

winterz:
please test Ken's fixes.
closing this pull request.
if you continue to experience problems please work with Ken to solve and create a new pull request.

champtar:
unable to crash libical,
thanks @ksmurchison

PR #221, opened 2015-09-24
icalrecur: export recurrence enum conversion routines

rsto:
The recurrence enumerations are exported in icalrecur.h, but their conversion routines are static to icalrecur.c. 

Since similar routines are exported for icalproperty, icalkind and others, I propose to also export the recurrence converters. That way, library users don't need to write their own (and possibly outdated) converters for recurrence enumerations.

All unit tests passed.


champtar:
+1, i need icalrecur_freq_to_string and icalrecur_weekday_to_string to compile pg_rrule

Issue #227, opened 2016-02-04
Fix libicu linkage in libical.pc


rossburton:
I'm cross-compiling libical in a sysroot and the installed libical.pc looks like this:

```
prefix=/usr
exec_prefix=${prefix}
libdir=${exec_prefix}/lib
includedir=${prefix}/include
threadslib=-lpthread
iculib=/data/poky-master/tmp-glibc/sysroots/intel-corei7-64/usr/lib/libicuuc.so /data/poky-master/tmp-glibc/sysroots/intel-corei7-64/usr/lib/libicui18n.so

Name: libical
Description: An implementation of basic iCAL protocols
Version: 2.0
Libs: -L${libdir} -lical -licalss -licalvcal ${threadslib} ${iculib}
Cflags: -I${includedir}
```

As you can see the iculib variable doesn't contain `-licuuc -licui18n` but the full paths to the libraries inside the sysroot.  Apart from being ugly, this breaks when the sysroot is moved.

winterz:
iculibdir=${SYSROOT}/usr/lib
iculib=-licuuc -licui18n
Libs: -L${libdir} -lical -licalss -licalvcal ${threadslib} -L${iculibdir} ${iculib}

any better? I'm not expert in cross-compiling.  not sure if there is a standard envvar for the sysroot

winterz:
ping? thoughts?

KrullKorg:
i don't know how cmake works so i cannot help, but i think it is better to use pkg-config, if the version of libicu provides it

winterz:
fixed in 1a41069c0dc054e59cd76fc3d8cf7cc2a78b5e8d

PR #230, opened 2016-04-20
Use of pkg-config to find libicu.

KrullKorg:
#227


winterz:
I'm not interested in this patch at this time, thanks anyway.

KrullKorg:
why? it adds a feature, how could not be interesting?

Issue #231, opened 2016-04-21
icaltime_compare unexpected behavior


stlaz:
Hi,

I noticed that when comparing two different times, one of which is floating, it's possible that these get identified as being the same time.

Example:

- 20160418T093957 in a time zone with +2 hours
- 20160418T073957, floating time

When fed to icaltime_compare, this first tries to convert both to UTC. The first one gets -2 hours, the second one stays the same, except it is now represented as a time in UTC, thus changing its meaning. This results in the times being considered the same, even though they certainly aren't. (Would also work the other way around.)

Also, icaltime_compare_with_zone does not exist and should probably be cleared from icaltime.h.

PR #232, opened 2016-04-25
Fixed comparison of float time with timezoned time

stlaz:
During comparison, float time would be converted to UTC if
compared to anything else resulting in wrong comparison as
float time should be time zone indenpendent.

Also removed non-existing function icaltime_compare_with_zone
from icaltime.h

https://github.com/libical/libical/issues/231

### DIFF
--- a/src/libical/icaltime.c
+++ b/src/libical/icaltime.c
@@ -778,8 +778,10 @@ int icaltime_compare(const struct icaltimetype a_in, const struct icaltimetype b
 {
     struct icaltimetype a, b;
 
-    /* We only need to perform time zone conversion if times aren't in the same time zone */
-    if (a_in.zone != b_in.zone || a_in.is_utc != b_in.is_utc) {
+    /* We only need to perform time zone conversion if times aren't in the same time zone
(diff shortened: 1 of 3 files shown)


stlaz:
BUMP, anything wrong with the patch?

winterz:
sorry, I missed this one.
the patch looks good, except the int_is check on line 1788 is now failing:
    int_is("icaltime_compare(): same UTC and NY",
           icaltime_compare(icttutc, icttny),
           0);

probably because of Daylight Savings time.
If I change
    icttny.hour = icttny.hour - 5;
to
   icttny.hour = icttny.hour - 4;
it works ok.

can you fix that problem?

stlaz:
That's actually weird as according to http://www.timeanddate.com/time/zone/usa/new-york, the 5-hour offset is right. Your system may be tainted. Anyway, added a "safer" date for this check so the test should not fail for anyone.

stlaz:
Did you find out what caused the issue at your side?

winterz:
committed with sha 3ab8fb407cf101c917ba7f98fecfcb148bc09851
thanks

PR #244, opened 2016-10-03
build: ICU must appear as Requires in pkgconfig

jengelh:
libical.pc specifies -licu-i18n in its Libs: field,
but no Requires: icu-i18n. As a result, the automatic dependency
generator in Linux distributions won't see the ICU requirement,
won't install it, and builds of secondary software fails.

$ gcc icalthing.c `pkg-config libical --cflags --libs`
[...]
gcc: error: /usr/lib64/libicuuc.so: No such file or directory

Move ICU from Libs to Requires, so the dependency scanner can do its
job. Specifically move it to Requires.private, since specifying -licu*
is not normally needed when libical is a shared library since it
already has it recorded in the ELF.

### DIFF
--- a/libical.pc.in
+++ b/libical.pc.in
@@ -2,11 +2,11 @@ prefix=@prefix@
 exec_prefix=@exec_prefix@
 libdir=@libdir@
 includedir=@includedir@
-threadslib=@PTHREAD_LIBS@
-iculibs=@ICUUC_LIBS@ @ICUI18N_LIBS@
 
 Name: libical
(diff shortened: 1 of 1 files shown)


winterz:
there is a problem though, if for some reason icu isnt' found.
so don't put the Requires line if libical is built without icu

winterz:
Thanks. I am closing this pull request since I just committed a slightly different solution in f70a26aaf836da0f9e242945151b49190d682d95

please make sure my fix works for you.
-Allen

PR #246, opened 2016-10-21
build: Pass 'CC' to the introspection scanner

ebassi:
This change avoids g-ir-scanner picking up a different compiler from the
environment than the one specified when building. When CC is absent from
the environment used to launch g-ir-scanner, it'll fall back to the
default behaviour of distutils, which is to use the compiler with with
distutils themselves were built. This causes compilation failure because
of missing system includes if distutils was built on a different machine
or with different arguments.

The behaviour introduced by this commit similar as the one provided by
GObject introspection itself in its autotools integration, except it's
missing CFLAGS, CPPFLAGS, and LDFLAGS, because their presence makes the
command line generated by cmake too long. The only way to match
behaviours would be to create a small wrapper script that set all the
environment variables and then called g-ir-scanner. Currently, this is
not necessary.

### DIFF
--- a/cmake/modules/GObjectIntrospectionMacros.cmake
+++ b/cmake/modules/GObjectIntrospectionMacros.cmake
@@ -50,7 +50,8 @@ macro(gir_add_introspections introspections_girs)
     set(_gir_libtool "--no-libtool")
 
     add_custom_command(
-      COMMAND ${INTROSPECTION_SCANNER}
+      COMMAND ${CMAKE_COMMAND} -E env "CC='${CMAKE_C_COMPILER}'"
+              ${INTROSPECTION_SCANNER}
               ${INTROSPECTION_SCANNER_ARGS}
               --namespace=${_gir_namespace}
               --nsversion=${_gir_version}


winterz:
unfortunately env isn't portable to Windows.  I'm asking Milan to review this patch.

ebassi:
As I said, the only other option is to have CMake generate a script file that:

 * sets the environment variables using the CMake variables
 * calls `g-ir-scanner` with the same arguments specified by the CMake custom command rule

The other reason why the environment variables are necessary is because they allow "cross-compilation" by wrapping the compiler under virtualization environments, like qemu.

The template for the invocation of `g-ir-scanner` comes from the `Makefile.introspection` file shipped by [gobject-introspection](https://git.gnome.org/browse/gobject-introspection/tree/Makefile.introspection) and installed alongside the rest of gobject-introspection; maybe the GObjectIntrospection module could call the rules on that file instead of generating its own targets.

ssssam:
CMake provides a 'command mode' to emulate certain shell commands in a portable way. So you can replace `env` with `cmake -E env`. Run `cmake -E help` for a synopsis.

ebassi:
@ssssam Yep, that's what the patch does: `COMMAND ${CMAKE_COMMAND} -E env ...`

The main issue, as I said in the commit message, is that I cannot put all the environment variables that `Makefile.introspection` uses, because it makes the command line too long.

Issue #264, opened 2016-12-14
Uncaught exception on OSX in test/regression-cxx.cpp


winterz:
see src/test/regression-cxx.cpp .  there is a conditional compile for !APPLE .  see if that's still needed and fix as necessary

winterz:
fixed now in commit cc04a4429414b0d612d78118442c289f6efdc412

PR #266, opened 2016-12-17
Set visibility of icalerrorenum to default

ksuther:
I only tested this on macOS, but the added #pragmas should be portable.

Fixes not being able to catch exceptions from libraries in clang (#264).


winterz:
tested.  works.  thanks!

PR #277, opened 2016-12-20
Generate Vala vapi file on the go

tintou:
This is required to use libical-glib with Vala, this is a soft dependency as it is only used if specified by the ICAL_GLIB_VAPI flag and only at required compilation time.


winterz:
we should also add a bullet to the ReleaseNotes.txt and maybe the Install.txt file.  can you look at doing these things?

tintou:
Okay, I'll do my best to provide it asap

tintou:
It's now in PR #279

Issue #256, opened 2016-12-03
Change -DUSE_INTEROPERABLE_VTIMEZONES default


winterz:
See the discussion at http://lists.infradead.org/pipermail/libical-devel/2016-January/000656.html

ksuther:
One practical reason to switch this default: It fixes expanding time zones past 2038 (otherwise icaltime_convert_to_zone will fail).

We've been running with USE_INTEROPERABLE_VTIMEZONES off because we have a separate layer that makes compact interoperable VTIMEZONE rules and didn't need it, but this fixes behavior internal to libical as well.

winterz:
just to be clear:  we will now use USE_INTEROPERABLE_VTIMEZONES=True by default.
any objections?

winterz:
Done in ff70a4ca002148c33dec543697e0403e5691e883

Issue #95, opened 2014-05-31
Regression: VTIMEZONE contains a lot of non-used timezone info, what mess with Google Calendar, EWS Calendar, etc ... [sf#95]


winterz:
I've bisect the code and I found the regression was inserted in  svn://svn.code.sf.net/p/freeassociation/code/trunk/libical@1130

For a clearer understanding of the problem, you also can find two calendars attached. The "expected_calendar.ics", was generated using latest libical with r1130 reverted; you can see it is populating VTIMEZONE only with used timezones. The "wrong_calendar.ics", on the other hand,  is populating VTIMEZONE with a lot of non-used timezones.

Also, you can find the patch reverting the r1130.

Please, let me know what kind of info I can provide to you to fix it.

Ah, moreover, two related bugs[0][1] were filled against Evolution (and its componentes), both can be fixed reverting r1130.

[0]: https://bugzilla.gnome.org/show_bug.cgi?id=708143
[1]: https://bugzilla.gnome.org/show_bug.cgi?id=710683

winterz:
fixed in 6e11823ec46ddff04c318034325da9631cd8e654
by default we use exact VTIMEZONEs. but you can now either at buildtime or at runtime opt to use inter-operable timezones

PR #298, opened 2017-05-27
Create VTIMEZONEs with RRULEs in icaltzutil_fetch_timezone

ksuther:
This is a follow-up to #256 and #95.

The change to have USE_INTEROPERABLE_VTIMEZONES on by default improves expansion of events in the future (past 2038), but it causes trouble expanding events in the past when generating VTIMEZONE components from system data (when not using built-in time zones).

The problem is that icaltzutil_fetch_timezone was creating a single VTIMEZONE component for the time zone back to 1970, which isn't necessarily accurate.

I've changed icaltzutil_fetch_timezone to generate VTIMEZONEs using RRULEs so that they're still interoperable but they're also accurate for the past (such as a date in 1980 in America/New_York, which is no longer using the same time zone rules).

I'm not quite done yet (the timezones test is failing), but is there any reason not to switch to something like this? It seems to combine the advantages of both modes, which means we can get rid of USE_INTEROPERABLE_VTIMEZONES completely.


winterz:
I'm happy to eliminate the INTEROPERABLE_VTIMEZONES stuff.
As for the actual implementation, I really don't have a comment since timezones confuse me.  if the tests all pass and nobody objects, then let's go for it.

ksuther:
Finished things up, the timezones test passes again.
The time zones generated should be more accurate for the past and future now.

The new implementation is sort of a hybrid of both approaches. The interoperable time zone code is still partially used, but it is run in a loop for every transition rather than taking just the last available transition.

Issue #293, opened 2017-03-24
Cleanup UsingLibical.txt 


winterz:
UsingLibical.txt could use a scrubbing and even better would be to incorporate it into the Doxygen manual and adding more info to the wiki.

xfbs:
Hey, I want to use libical for a project and saw that the Doxygen documentation is a bit sparse. Maybe I could help clean it up?

winterz:
absolutely. see issue #175
patches are happily accepted.

xfbs:
I'll send my pull requests your way.

winterz:
UsingLibical has been converted to markdown and is now part of the Manual (book) and linked to from the API docs.

done

PR #302, opened 2017-07-11
Cleanup UsingLibical.txt

xfbs:
In response to #293, I played with the UsingLibical.txt a bit and ended up converting it to Markdown (since it's more accessible, plus you get links, syntax highlighting and tables for free while still being legible). I also fixed a fix mistakes here and there, reformatted the example code just a little bit. I found that some of the examples don't seem to work — I'll have a look at those later.


ksuther:
Thanks! Markdown is definitely much better choice for that file.

winterz:
nice! Thanks

PR #303, opened 2017-07-12
Adds doxygen docs for icalerror.h

xfbs:
This adds doxygen documentation for `libical/icalerror.h` — should be fairly complete.

### DIFF
--- a/src/libical/icalerror.h
+++ b/src/libical/icalerror.h
@@ -30,12 +30,12 @@
  * @file icalerror.h
  * @brief Error handling for libical
  *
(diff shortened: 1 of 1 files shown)


winterz:
Awesome stuff,
could you tell your editor not to leave trailing whitespace ?

also I just noticed you misspelled "occured".. I'll fix

xfbs:
Sure, I'll look into the trailing whitespace stuff — I'm sure there is some option or plugin for vim to highlight things like that or automagically remove it. Thanks for letting me know! Might as well install a spellchecker while I'm at it.

PR #323, opened 2017-12-29
Make libical reprodicuble

jelly:
Sort the hash which generates the C struct so that the struct is
identical on every build, since normally this is random. With this
change libical is fully reprodicuble.

See the non-reprodicble output here https://tests.reproducible-builds.org/archlinux/extra/libical/libical-3.0.1-1-x86_64.pkg.tar.xz.html


winterz:
committed with bcd67ef02dec59b7261f045ad6a2c0bc29fd620a
thanks for the patch

PR #324, opened 2018-02-04
Please make the typedef & include generation reproducible.

lamby:
Whilst working on the Reproducible Builds effort [0], we noticed
that libical3 could not be built reproducibly as it iterates over
the filesystem as well as iterates over the keys of a hash, both
of which are non-deterministic.

Patch attached.

 [0] https://reproducible-builds.org/

Filed in Debian as https://bugs.debian.org/889565

### DIFF
--- a/src/libical-glib/tools/generator.c
+++ b/src/libical-glib/tools/generator.c
@@ -1096,6 +1096,7 @@ void generate_forward_declarations_header_file(GList *structures)
     gchar *typeName;
     gchar *typeKind;
     GList *link;
+    GList *typeNamesList = NULL;
     GHashTable *typeNames;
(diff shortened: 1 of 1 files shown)


winterz:
unfortunately travis-ci is showing segfaults
https://travis-ci.org/libical/libical/jobs/337225459, for example

lamby:
Ah, can't reproduce that locally but by-inspection I think it was that `filenames` could be uninitialised. Trying again..

lamby:
@winterz There you go :)

winterz:
commited in sha 6659ae7d5d4289068c5a61bb27b885a4fc600ca6

lamby:
Thanks :)

PR #326, opened 2018-02-19
Fix for Issue #325 - icalcomponent_foreach_recurrence

### LINKED ISSUE #325: icalcomponent_foreach_recurrence omits first occurrence if DTSTART doesn't confirm to RRULE.
If the DTSTART in a VEVENT doesn't conform to the contained RRULE, icalcomponent_foreach_recurrence might omit the first recurrence instance.

I know, that this case is specified as undefined by the RFC, but in real life DTSTART often is a value provided by end users and might not fully conform to the RFC in this case. We could use libraries to sanitize such values but then we need libraries that handle the case in a meaningful way.

The problem can be reproduced with following code:

```C
void reproduce_libical_first_recurrence_missing_callback(icalcomponent *comp, struct icaltime_span *span, void *data) {

	int* foundExpected = (int*)data;
	(*foundExpected)++;
}

void reproduce_libical_first_recurrence_missing() {

	const char* calStr =
		"BEGIN:VCALENDAR\n"
		"BEGIN:VEVENT\n"
		// 14th is not a tuesday, so it doesn't conform to the rule.
		"DTSTART;20180214T020000Z\n"
		"DURATION:PT4H\n"
		"RRULE:FREQ=WEEKLY;COUNT=2;BYDAY=TU\n"
		"END:VEVENT\n"
		"END:VCALENDAR\n";

	icalcomponent *calendar = icalparser_parse_string(calStr);
	icalcomponent *event = icalcomponent_get_first_component(calendar, ICAL_VEVENT_COMPONENT);

	struct icaltimetype t_start = icaltime_from_string("20180219T000000Z");
	struct icaltimetype t_end = icaltime_from_string("20180221T000000Z");

	int foundExpectedCnt = 0;
	// we expect exactly one callback, namely on tu, 20th.
	icalcomponent_foreach_recurrence(event, t_start, t_end, reproduce_libical_first_recurrence_missing_callback, &foundExpectedCnt);

	icalcomponent_free(calendar);

	if (foundExpectedCnt == 1)
		printf("ok");
	else
		printf("test failed - first recurrence instance wasn't returned.");
}
```

This seems to be similar to but distinct from #291 .

[Edit] Fixed format of DURATION property, fixed comment and removed obsolete printf - doesn't change test outcome.

minichma:
This PR fixes #325, i.e. function icalcomponent_foreach_recurrence(). Basicly the function always expected a VEVENT's DTSTART to be the first occurrence and treated this case separately by explicitely returning this instance to the caller. When later in the function the RRULE was iterated, the first instance (which was expected to be the DTSTART) was skipped. But in cases where a RRULE was present and the DTSTART didn't conform to the rule, the assumption was wrong and not the already returned DTSTART was skipped (which wasn't returned by the RRULE's iterator) but the first correct recurrence instance. This is fixed with this branch.

Several test cases have been added to the regression tests that should cover the most important cases.

### DIFF
--- a/src/libical/icalcomponent.c
+++ b/src/libical/icalcomponent.c
@@ -924,9 +924,12 @@ void icalcomponent_foreach_recurrence(icalcomponent *comp,
     limit_span.start = limit_start;
     limit_span.end = limit_end;
 
-    /* Do the callback for the initial DTSTART entry */
+    rrule = icalcomponent_get_first_property(comp, ICAL_RRULE_PROPERTY);
 
-    if (!icalproperty_recurrence_is_excluded(comp, &dtstart, &dtstart)) {
+    /* If there is no RRULE present (in which case DTSTART is the only returned instance),
+       do the callback for the initial DTSTART entry. */
+
+    if ((rrule == NULL) && !icalproperty_recurrence_is_excluded(comp, &dtstart, &dtstart)) {
     /** call callback action **/
         if (icaltime_span_overlaps(&basespan, &limit_span))
             (*callback) (comp, &basespan, callback_data);
@@ -936,17 +939,12 @@ void icalcomponent_foreach_recurrence(icalcomponent *comp,
     dtduration = basespan.end - basespan.start;
 
     /* Now cycle through the rrule entries */
-    for (rrule = icalcomponent_get_first_property(comp, ICAL_RRULE_PROPERTY);
+    for (;
          rrule != NULL; rrule = icalcomponent_get_next_property(comp, ICAL_RRULE_PROPERTY)) {
 
         struct icalrecurrencetype recur = icalproperty_get_rrule(rrule);
         icalrecur_iterator *rrule_itr = icalrecur_iterator_new(recur, dtstart);
         struct icaltimetype rrule_time;
-
-        if (rrule_itr)
-            rrule_time = icalrecur_iterator_next(rrule_itr);
-    /** note that icalrecur_iterator_next always returns dtstart
-        the first time.. **/
 
         while (rrule_itr) {
             rrule_time = icalrecur_iterator_next(rrule_itr);
--- a/src/test/regression.c
+++ b/src/test/regression.c
@@ -530,6 +530,150 @@ void test_components()
 	icalerror_set_errors_are_fatal(estate);
 }
 
+static void test_component_foreach_callback(icalcomponent *comp, struct icaltime_span *span, void *data) {
+
+    int* foundExpected = (int*)data;
+    (*foundExpected)++;
+}
+
+int test_component_foreach_parameterized(int startOffsSec, int endOffsSec, int expectedFoundInstances) {
+
+    const char* calStr =
+        "BEGIN:VCALENDAR\n"
+        "BEGIN:VEVENT\n"
+        "DTSTART;20180220T020000Z\n"
+        "DURATION:PT1H\n"
+        "RRULE:FREQ=WEEKLY;BYDAY=TU\n"
+        "END:VEVENT\n"
+        "END:VCALENDAR\n";
+
+    icalcomponent *calendar = icalparser_parse_string(calStr);
+    icalcomponent *event = icalcomponent_get_first_component(calendar, ICAL_VEVENT_COMPONENT);
+
+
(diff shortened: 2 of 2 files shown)


src/libical/icalcomponent.c
```diff
@@ -936,18 +939,13 @@ void icalcomponent_foreach_recurrence(icalcomponent *comp,
     dtduration = basespan.end - basespan.start;
 
     /* Now cycle through the rrule entries */
-    for (rrule = icalcomponent_get_first_property(comp, ICAL_RRULE_PROPERTY);
+    for (;
```
sebastianhaas: Is this line break intentional?

src/test/regression.c
```diff
@@ -530,6 +530,150 @@ void test_components()
...
+        "END:VCALENDAR\n";
+
+    icalcomponent *calendar = icalparser_parse_string(calStr);
+    icalcomponent *event = icalcomponent_get_first_component(calendar, ICAL_VEVENT_COMPONENT);
+
+
+    struct icaltimetype dtstart = icaltime_from_string("20180220T020000Z");
+    struct icaltimetype it_start = dtstart;
+    struct icaltimetype it_end = dtstart;
+	int foundCnt = 0;
```
sebastianhaas: Probably unintended indentation

src/test/regression.c
```diff
@@ -530,6 +530,150 @@ void test_components()
...
+        "END:VCALENDAR\n";
+
+    icalcomponent *calendar = icalparser_parse_string(calStr);
+    icalcomponent *event = icalcomponent_get_first_component(calendar, ICAL_VEVENT_COMPONENT);
+
+    struct icaltimetype t_start = icaltime_from_string("20180219T000000Z");
+    struct icaltimetype t_end = icaltime_from_string("20180221T000000Z");
+
+    int foundExpectedCnt = 0;
+	int i;
```
sebastianhaas: Probably unintended indentation

src/libical/icalcomponent.c
minichma: The line break was left as before to reduce the size of the code change.

src/test/regression.c
minichma: Yes (tab vs spaces), was introduced when last-minute fixing the pedantic compiler errors. Will fix this.

src/test/regression.c
minichma: Yes, see above.

ksmurchison:
This doesn't look correct.  The first instance of an RRULE (e.g. the DTSTART) can be excluded by an EXDATE.  That case appears to have been removed by this patch.

Is the intent to treat a DTSTART that doesn't match the RRULE pattern as an RDATE?  What if the DTSTART is a nonsensical value, e.g. 20180230?  RFC5545 says that nonsensical values generated by RRULE expansion are skipped, should we also apply that to DTSTART?

minichma:
@ksmurchison: I think I don't fully understand your comment. In fact DTSTART can still be excluded by an EXDATE as before. Can you provide an example of where things have worsen?

> Is the intent to treat a DTSTART that doesn't match the RRULE pattern as an RDATE?

No, the intent is rather not to include such instances in the recurrence set at all. The RFC leaves this undefined, but I think, it's a pretty common case that recurrence rules are thought to start after DTSTART. E.g. one could schedule an event happening every friday, starting on next friday. "Next friday" could for simplicity be specified by a DTSTART with today's date, where today could be any weekday. If one would like to have DTSTART itself in the recurrence set, it could easily be included as RDATE. Removing a value from the recurrence set is much harder. As the RFC (unfortunately) leaves this undefined, the proposed implementation seems to be the most natural one to me. So the idea of the fix is simply to include the DTSTART in the recurrence set if there is no RRULE but let the RRULE generate the instances if present. As stated in the RFC, the generated set SHOULD include DTSTART either way.

> What if the DTSTART is a nonsensical value, e.g. 20180230? RFC5545 says that nonsensical values generated by RRULE expansion are skipped, should we also apply that to DTSTART?

I'm not sure, I understan this correctly. I don't think, we have to handle this case here, do we? At least I think, behaviour hasn't changed with this PR. Please provide a more detailed example if I'm wrong.

ksmurchison:
I misunderstood the actual problem.  This patch does look correct.  I will probably do a little code cleanup afterwards.

PR #331, opened 2018-03-06
make memory management configurable

minichma:
This PR adds functionality that gives users more control over memory management within libical. Today malloc, realloc and free are used to allocate and deallocate memory but this might not always be suitable, especially in resource-constrained environments. The possibility of replacing these functions also allows implementing test cases that test for the correctness and robusness of memory management. The `icalmemory_new_buffer()` and related functions have already been present in preparation for these scenarios but have not yet been configurable and not been used consistently. This PR replaces calls to malloc, realloc, free and strdup with the corresponding preexisting wrapper functions throughout the libical core component. It makes the wrapper functions themselfes configurable, so custom memory management functions can be used instead of malloc & co.

## Test
This PR extends the regression tests with some basic functionality that tests for consistency and robusnes against out of memory errors. Regarding consistency, the extension ensures, that memory allocated with `icalmemory_new_buffer()` is not freed with the standard system `free()` function and that on the other hand `icalmemory_free_buffer()` is only used to free memory that has been allocated using `icalmemory_new_buffer()`. No tests have been added that test for memory leaks (i.e. verifying that all memory that has been allocated is freed too) but adding such tests would be trivial with the functionality introduced by this PR. However, it would require consideration of how to handle cached components like timezones that are not necessarily freed at the end of each test and would therefore look like memory leaks to naive test cases.

Tests have been added that make sure, that out of memory errors (where `icalmemory_new_buffer()` returns NULL) are handled gracefully in certain situations. These tests cover only a limited set of functionality and brought up some issues in icalarray and icalrecur that were fixed as part of this PR. Experiments with e.g. the icalparser component (which are not included in this PR) showed little robustness against out of memory errors there. Making all components robust against out of memory errors would be improtant for use especially in resource-constrained environments but is out of the scope of this PR and should be considered for future work.

## Compatibility
This change is fully backwards compatible because by default still the standard malloc, realloc, free functions are used, just like before. So users can free() memory just as before, even if the freed memory was allocated using libcalmemory_new_buffer. However, the change is not fully forwards compatible once people start using `icalmemory_set_mem_alloc_funcs()` if the use of icalmemory is introduced in more components. I therefore added a note to `icalmemory_set_mem_alloc_funcs()`, discouraging its use if using more than just the core libical component until introduction in all components has been completed.

## CMAKE Flag ICAL_USE_MALLOC
The CMAKE flag `ICAL_USE_MALLOC` was introduced, controlling whether or not the standard system memory allocation functions should be used by default. It makes sense to have this explicit flag rather than introducing something like a `HAVING_MALLOC` flag (which would only indicate whether malloc is present) because it can make sense to avoid linking malloc, etc., even if they are present. This is especially true in resource constrained environments where linking malloc would inacceptably increase the size of the compiled binary.

## Future work
* Use libcalmemory_new_buffer throughout other components (libicalvcal, etc.) too.
* Produce compile errors when using malloc, etc. withing components where libcalmemory_new_buffer should be used to avoid unintended reintroduction of former functions. This could be achieved by undefining malloc, etc.
* Add more tests that test for out of memory errors (libcalmemory_new_buffer returns NULL) similar to the tests introduced with this PR. However, such tests will bring up issues that need to be fixed in order to succeed (e.g. in the icalparser component).
* Add more tests that test for memory leaks.

[Edit] No tests for memory leaks.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -18,6 +18,11 @@
 #  Set to prevent empty properties from being replaced with X-LIC-ERROR properties.
 #  Default=false
 #
+# -DICAL_USE_MALLOC=[true|false]
+#  If set, the standard system functions (malloc, realloc, free) will be used for memory management by default.
+#  If not set, the user must use icalmemory_set_mem_alloc_funcs function to provide the functions to use.
+#  Default=true (use system malloc, etc. by default).
+#
 # -DUSE_BUILTIN_TZDATA=[true|false]
 #  Set to build using our own timezone data.
 #  Default=false (use the system timezone data on non-Windows systems)
@@ -263,6 +268,18 @@ if(ICAL_ALLOW_EMPTY_PROPERTIES)
   set(ICAL_ALLOW_EMPTY_PROPERTIES 1)
 else()
   set(ICAL_ALLOW_EMPTY_PROPERTIES 0)
+endif()
+
+option(ICAL_USE_MALLOC "Make use of the standard system memory management functions (malloc, realloc, free)." True)
+add_feature_info(
+  "Option ICAL_USE_MALLOC"
+  ICAL_USE_MALLOC
+  "the standard system memory management functions will be used by default"
+)
+if(ICAL_USE_MALLOC)
+  set(ICAL_USE_MALLOC 1)
+else()
+  set(ICAL_USE_MALLOC 0)
 endif()
 
 option(USE_BUILTIN_TZDATA "build using our own timezone data, else use the system timezone data on non-Windows systems. ALWAYS true on Windows.")
--- a/config.h.cmake
+++ b/config.h.cmake
@@ -175,6 +175,10 @@
 /* Define to prevent empty properties from being replaced with X-LIC-ERROR properties */
 #define ICAL_ALLOW_EMPTY_PROPERTIES ${ICAL_ALLOW_EMPTY_PROPERTIES}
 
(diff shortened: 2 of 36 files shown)


ksmurchison:
Why do we need  compile-time option?  Shouldn't malloc/realloc/free be the default functions unless overridden with the new API call?

minichma:
The compile-time option allows avoiding linking these functions, which can be necessary on embedded devices.

winterz:
I'm going to commit this to a new branch and take a closer look and we can play with it.  this is one of those patches where just doing a code review doesn't feel like enough.

winterz:
see the mem-manage branch.
@minichma I will give invite you to join so you can commit yourself from now on.
-Allen

minichma:
Great, thanks Allen!

Am 14.03.2018 22:26 schrieb "Allen Winter" <notifications@github.com>:

see the mem-manage branch.
@minichma <https://github.com/minichma> I will give invite you to join so
you can commit yourself from now on.
-Allen

—
You are receiving this because you were mentioned.

Reply to this email directly, view it on GitHub
<https://github.com/libical/libical/pull/331#issuecomment-373180347>, or mute
the thread

minichma:
I tried to merge a minor fix into the mem-manage branch, but don't seem to have permission to do so.

winterz:
can you try again please

minichma:
Now it works, thanks!

PR #332, opened 2018-03-12
Improve memory efficiency of icalarray_sort()

minichma:
This PR improves the efficiency and required RAM when sorting icalarrays. The function `icalarray_sort()` used to use the stdlib's qsort function for sorting arrays. As qsort requires the array to be stored as an contiguous block of memory, which isn't the case in icalarray, `icalarray_sort()` had to create a temporary array, big enough to hold all of the array's entries, and copying the whole content back and forth. In resource constrained environment, the additional required memory can have significant impact.

This PR replaces the stdlib's qsort with a more generic implementation that allows sorting arrays, that are not stored in a contiguous block of memory, thus allowing sorting the array in place and avoiding the cost of additional temporary memory and associated copying.

## Copyright/License
The qsort algorithm introduced with this PR is based on code from the [PDCLib Project](https://bitbucket.org/pdclib/pdclib/raw/c8dc861df697a6c8bddbcbf331d9b6fcae6e2f4d/functions/stdlib/qsort.c), which is placed under the Creative Commons Zero license, which allows us to "use PDCLib, in whole or in part, as you choose, with no restrictions". Including code in libical that is based on this license shouldn't have any negative influence on libical's license. To conform to libical's practice, I added a copyright header in the modified code, stating that the modified code is placed under LGPL or MPL, while citing the original license statement. To my understanding, this approach should be in line with both, PDCLib's and libical's licensing.


minichma:
Rebased this branch on current master, to avoid dependency on mem-manage branch.

minichma:
@winterz Does this PR have any chance of getting reviewed and merged? If yes, I would rebase it (again) to fix the conflicts but I'd want to avoid the work if the change is not appreciated.

winterz:
sorry for not responding on this one.
I've been putting my libical time onto getting the 3.0.5 released.

Yes if you could fix the conflicts I'll see about merging to master.
Thank you for your work, I'm sorry I don't have that much time to put on libical these days.

minichma:
Oh, no worries! I appreciate the huge amount of work you guys did and do a
lot. Will try to fix this in the next couple of days.

Allen Winter <notifications@github.com> schrieb am Mi., 1. Mai 2019, 22:25:

minichma:
@winterz I rebased and cleaned the branch - it's ready for review now. I suggest to review the code commit by commit. The very first commit only contains code from the PDCLib project, which is modified in the following commits. I added tests for the `icalarray_sort` function, which implicitly also tests the new `qsort_gen()`.  Testing the new `qsort_gen()` individually is not easily possible via `regression.c`, because the new functions are not exposed via the public interface.

winterz:
committed with sha d101b1798dfab5aa6730e63424156453f859f114
thanks!

PR #337, opened 2018-03-29
Itroduce icalrecur_iterator_set_end() to allow for performance otimizations

minichma:
This PR introduces the icalrecur_iterator_set_end() function, which allows to set an end date for a recurrence iterator. This can make sense when iteration is only required up to a specific date/time because it allows avoiding additional iterator-internal coputational efforts in such cases.

### DIFF
--- a/src/libical/icalrecur.c
+++ b/src/libical/icalrecur.c
@@ -829,6 +829,7 @@ struct icalrecur_iterator_impl
 
     struct icaltimetype rstart;      /* DTSTART in RSCALE  */
     struct icaltimetype istart;      /* Gregorian start time for iterator */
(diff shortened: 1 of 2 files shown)


ksmurchison:
I'm confused as to why the iterator needs to do this.  Why can't the application simply compare the occurrence that is calculated against its desired end date?

minichma:
The idea is to improve performance and, even more importantly, make runtime more predictable, by avoiding potential extra iterations internal to libical.

minichma:
@winterz: Regarding ReleaseNotes.txt - shall I modify the file within this branch? Which release?

winterz:
master only for now.  I may backport to the 3.0 branch at some point.

Issue #339, opened 2018-04-05
libicalvcal fails parsing VCF files produced by outlook


bosim:
Outlook produces VCF files like these:
```
BEGIN:VCARD
VERSION:2.1
N:de Tester 04-04;Test;;;(outlook)
FN:Test de Tester 04-04 (outlook)
ORG:Test BV
TITLE:Tester
TEL;WORK;VOICE:(010) 123-4567
TEL;HOME;VOICE:(010) 123-4568
TEL;CELL;VOICE:(010) 123-4560
ADR;WORK:;;Workstreet 99;Worktown;;1111AA;Netherlands
LABEL;WORK;ENCODING=QUOTED-PRINTABLE:Workstreet 99=0D=0A=
1111AA  Worktown=0D=0A=
Netherlands
ADR;HOME;PREF:;;Homestreet 99;Hometown;;1111ZZZ;Netherlands
LABEL;HOME;PREF;ENCODING=QUOTED-PRINTABLE:Homestreet 99=0D=0A=
1111ZZZ  Hometown=0D=0A=
Netherlands
X-MS-OL-DEFAULT-POSTAL-ADDRESS:1
URL;HOME:https://www.personalhomepage.com
URL;WORK:https://www.businesshomepage.com
EMAIL;PREF;INTERNET:tester@test.com
EMAIL;INTERNET:tester2@test.com
EMAIL;INTERNET:tester3@test.com
X-MS-CARDPICTURE;TYPE=JPEG;ENCODING=BASE64:
 /9j/4AAQSkZJRgABAQEAkACQAAD/2wBDAAcFBQYFBAcGBQYIBwcIChELCgkJChUPEAwRGBUa
 GRgVGBcbHichGx0lHRcYIi4iJSgpKywrGiAvMy8qMicqKyr/2wBDAQcICAoJChQLCxQqHBgc
 KioqKioqKioqKioqKioqKioqKioqKioqKioqKioqKioqKioqKioqKioqKioqKioqKir/wAAR
 CACUACcDASIAAhEBAxEB/8QAHwAAAQUBAQEBAQEAAAAAAAAAAAECAwQFBgcICQoL/8QAtRAA
 AgEDAwIEAwUFBAQAAAF9AQIDAAQRBRIhMUEGE1FhByJxFDKBkaEII0KxwRVS0fAkM2JyggkK
[...]

PR #340, opened 2018-04-06
Fix parsing :<CR>?<LF>

bosim:
Here is a mixed bag of changes to vcc.y to get it running using ~~```bison -y vcc.y --output vcc.c```~~ ```byacc -p "mime_" -o vcc.c vcc.y``` which seems to be the original compiler-compiler (used it instead of bison to get a minimal diff). Wonder why these changes are necessary? Nevertheless, removed the comments that will enable skipping the linefeeds after colon, and it fixes the problem described in issue #339, however I am unsure why it was commented out in the first place.

### DIFF
--- a/src/libicalvcal/vcc.c
+++ b/src/libicalvcal/vcc.c
@@ -1,24 +1,103 @@
-#ifdef HAVE_CONFIG_H
-#include <config.h>
-#endif
-
-#include "vcc.h"
-#include <ctype.h>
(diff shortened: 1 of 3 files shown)


bosim:
@winterz what do you think?

winterz:
@bosim I'm sorry I haven't had a chance to look very closely at this pull request. I'll try to make time this week.  Maybe another one of our project maintainers has time?

winterz:
merged into master

PR #346, opened 2018-05-11
icalcomponent.h: rewrite comment for dtend/duration API

ohwgiles:
The previous comment inaccurately stated that the set_duration
and set_dtend functions would each calculate the change to
the other property


winterz:
committed 9e68f7eba6941246630893004fa5d10d4b211da0
thanks

PR #356, opened 2018-09-25
GObject Introspection - do not change the .gir and .typelib name on every minor/micro version

tintou:
Otherwise it means that every application using the bindings has to target a very specific version.
The $MAJOR-0 version is the current standard for this (see GLib, GTK…).


tintou:
@winterz Can you review this? It's very difficult to use the Python bindings in the current state and also difficult to package the successive versions of libical in Debian without this patch.

winterz:
I have no idea what to make of this.  I don't know anything about GObject Introspection nor packaging requirements.  

Would you want this in the 3.0.x series as well as master?
I'll ask Milan to review the patch.

PR #359, opened 2018-10-30
Update juldat() and caldat() functions to avoid floating point types.

minichma:
This PR updates the juldat() and caldat() functions to avoid the use of floating point types. It also removes functionality that is not used in context of libical (i.e. time-related computations). This improves portability to (i.e. embedded, resource constrained) platforms not supporting FP-arithmetics natively, e.g. ARM Cortext M. The change is intended to improve performance and reduce code size.

## Test
Tests have been added that compare the new implementation against the old one. It compares the application to some selected border case values and randomly selected ones.

### DIFF
--- a/src/libical/astime.h
+++ b/src/libical/astime.h
@@ -1,3 +1,7 @@
+/* 
+ * This work is based on work from Hiram Clawson and has been modified to the
+ * needs of the libical project. The original copyright notice is as follows:
+ */
 /*
  *      Copyright (c) 1986-2000, Hiram Clawson
  *      All rights reserved.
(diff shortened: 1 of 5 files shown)


src/libical/astime.h
```diff
@@ -46,24 +71,18 @@
...
-    double d_minute;            /**<    [0.0-59.9999] includes second   */
-    double d_second;            /**<    [0.0-59.9999]   */
     int weekday;                /**<    [0-6]   */
     int day_of_year;            /**<    [1-366] */
 } UTinstant, *UTinstantPtr;
 
 /*      Functions in caldate.c  */
 
-LIBICAL_ICAL_EXPORT long caldat(UTinstantPtr);      /** converts julian date to year,mo,da */
+LIBICAL_ICAL_EXPORT void caldat(UTinstantPtr);    /** converts julian date to year,mo,da */
```
winterz: I don't want to break binary and source compatibility.
we can add new new functions and deprecate the old one, however.

so I guess let's continue to return a value or we need a function.

src/libical/astime.h
```diff
@@ -46,24 +71,18 @@
...
 
-LIBICAL_ICAL_EXPORT double juldat(UTinstantPtr);    /** returns julian day from year,mo,da */
+LIBICAL_ICAL_EXPORT void juldat(UTinstantPtr);    /** returns julian day from year,mo,da */
```
winterz: same here.

src/libical/astime.h
winterz: right.  if we don't need to export the new ones (hopefully not) then let's not.

winterz: put the doxygen comment like the following in astime.h immediately above the function declaration.  for example:
/** Converts Julian date to year,mo, da
 ** @deprecated use caldat_int() instead 
*/
 LIBICAL_ICAL_EXPORT long caldat(UTinstantPtr); 

Also let's mention the deprecated functions in the top-level ReleaseNotes.txt.
Do we want this in a 3.0.5 release or is master (3.1.0) good enought?

minichma: Updated as suggested. Not sure, whether the new funcs should be mentioned in the deprecation message, as they are considered internal now (not exported). Still mentioned them to clearify internal use.

I think, 3.1.0 should be fine.

PR #364, opened 2019-01-04
Consistency between const and non-const params for clone functions

EmielBruijntjes:
The icalvalue_new_clone() function expects it's parameter to be const. This makes sense, because the to-be-cloned value is not going to be altered. All the other *_new_clone() functions in this library however, have a non-const parameter, even though they are not going to alter their argument either. This unnecessary inconsistency make it harder to create a common clone and destruct signature when you try to make a common C++ template class for all these types. This pull requests fixes the inconsistency.


winterz:
problem: this change isn't source or binary compatible.

I think I'd like to keep source compatibility at least; so:
1. deprecate the existing signatures (see the_deprecated macro)
2.  rename the new clone functions from icalwhatever_new_clone() to icalwhatever_new_constclone()
3. document this in the ReleaseNotes.txt

see 3cec913271b37ab98490e65520d198d7ecda62a3 for how we did something similar for caldat() and juldat()

note to self: for master only . do not cherry-pick to the 3.0 branch

EmielBruijntjes:
Hello Allen,

This is a decision for you to make of course. Would you like me to pick it up? To be honest, I do not really like the proposed naming "icalwhatever_new_constclone" because it suggests that the const-ness of the parameter is a _special case_, while in reality it is the normal case (in fact: to me the name even seems to imply that the _returned clone_ is going to be const).

May I suggest to use the name "icalwhatever_clone()" for the clone-functions, and make the icalwhatever_new_clone() calls all deprecated?

winterz:
> Hello Allen,
> 
> This is a decision for you to make of course. Would you like me to pick it up? To be honest, I do not really like the proposed naming "icalwhatever_new_constclone" because it suggests that the const-ness of the parameter is a _special case_, while in reality it is the normal case (in fact: to me the name even seems to imply that the _returned clone_ is going to be const).
> 
> May I suggest to use the name "icalwhatever_clone()" for the clone-functions, and make the icalwhatever_new_clone() calls all deprecated?

Yes, that's a much better name.  Please do.  Happy to have more help.
-Allen

winterz:
Scanning dependencies of target ical_cxx
[ 51%] Building CXX object src/libical/CMakeFiles/ical_cxx.dir/icalparameter_cxx.cpp.obj
icalparameter_cxx.cpp
c:\projects\libical\src\libical\icalparameter.h(103) : error C4430: missing type specifier - int assumed. Note: C++ does not support default-int

EmielBruijntjes:
Yes, I am looking at it now. The _deprecated macro does not seem to work for C++. Maybe you already know what I'm doing wrong? Otherwise I'll dig a little deeper.

Some things that I'm also not sure about: the changes that I made to the xml files in src/libical-glib. I just added extra methods for the *_clone() functions and kept the *_new_clone() functions too. Is this the right way to go? Is there a way to officially mark things as deprecated?

And I have not yet touched things inside the Net-ICal-Libical directory. I'm not so sure what this code is for, and whether it should be changed, and how.

EmielBruijntjes:
I hope you don't mind, but I took the freedom to rename the _deprecated macro to LIBICAL_DEPRECATED. This prevents possible conflicts with other libraries that may also have a _deprecated function or macro. It also respects the conventions that macro's are almost always in upper-case, and that they are prefixed with the name of the library that defined them.

I've also changed the new_clone() naming in the wrapper-libraries for perl and glib, but kept the old naming for backwards compatibility. I thought that since the underlying C library is going to use different naming, the wrapper libraries will probably want to follow this too. I did my best to filter out all "new_clone" calls in the code and replace them with a call to "clone", while at the same time keeping the old functions around for compatibility with existing code.

If there is anything else that I can do, please let me know.

winterz:
thank you Emile.
Committed with 7dc9a9674120fa8be2fc33e627211b515475207f

I made some small fixes

PR #366, opened 2019-01-09
Fixed incorrect handling of a weekly RRULE at the beginning of the year

ksuther:
Found an edge case for weekly recurrences at the beginning of the year. Regression test included.

---

This RRULE on January 1 would result in the second occurrence being January 7 rather than January 2.

FREQ=WEEKLY;WKST=SU;INTERVAL=2;BYDAY=MO,TU,WE,TH,FR

The problem is that day_diff wasn't handling cases where a > b and would count days of the year incorrectly. Temporarily swaping a and b and then performing the day_diff calculation avoids the issue.

### DIFF
--- a/src/libical/icalrecur.c
+++ b/src/libical/icalrecur.c
@@ -1959,6 +1959,17 @@ static int __day_diff(icalrecur_iterator *impl, icaltimetype a, icaltimetype b)
         diff = get_day_of_year(impl, b.year, b.month, b.day, NULL) -
             get_day_of_year(impl, a.year, a.month, a.day, NULL);
     } else {
+        /* Swap a and b if a is greater than b */
+        int flipped = 0;
+
+        if (a.year > b.year) {
+            icaltimetype temp = a;
+
+            a = b;
+            b = temp;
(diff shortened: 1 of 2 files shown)


src/libical/icalrecur.c
```diff
@@ -1959,13 +1959,29 @@ static int __day_diff(icalrecur_iterator *impl, icaltimetype a, icaltimetype b)
         diff = get_day_of_year(impl, b.year, b.month, b.day, NULL) -
             get_day_of_year(impl, a.year, a.month, a.day, NULL);
     } else {
+        /* Swap a and b if a is greater than b */
+        int flipped = 0;
```
winterz: in order to make MSVC happy, you probably have to move the declaration for 'int flipped' near the top of the function.

ksmurchison:
Good catch Kent!  This fix works for me.

ksuther:
I think `flipped` is fine, it's that `year` ended up getting pushed down. I'll see if that builds on MSVC 2012 now.

ksuther:
Should be better now.

PR #368, opened 2019-01-16
Apply changes found during Evolution Data Server port to libical-glib

tintou:
A series of patches found by Milan Crha (mcrha).
The description of each change is written in the commit message directly.

### DIFF
--- a/src/libical-glib/api/i-cal-array.xml
+++ b/src/libical-glib/api/i-cal-array.xml
@@ -13,7 +13,7 @@
   You should have received a copy of the GNU Lesser General Public License
   along with this library. If not, see <http://www.gnu.org/licenses/>.
 -->
- <structure namespace="ICal" name="Array" native="icalarray" destroy_func="icalarray_free">
+ <structure namespace="ICal" name="Array" native="icalarray" is_possible_global="true" destroy_func="icalarray_free">
 	<method name="i_cal_array_size" corresponds="CUSTOM" kind="other" since="1.0">
 		<parameter type="ICalArray *" name="array" comment="The #ICalArray"/>
 		<returns type="gint" comment="The size of current array."/>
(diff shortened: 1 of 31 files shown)


winterz:
do we want this in the 3.0 branch for a possible 3.0.5 release?

tintou:
From Milan:
That would be ideal, even it changes libical-glib API. On the other hand, I'd guess it's not used by anyone yet. An interesting approach would be to add something like LIBICAL_GLIB_UNSTABLE_API define as a requirement for the time being, to express that things can change. At least until the main consumer (evolution-data-server, possibly with its clients) is ported. Definitely do not hurry with the 3.0.5 just because of these changes.

winterz:
How about I give you commit privileges so you can do whatever you need for the libical-glib stuff?

My only concern would be binary and source compatibility.  I'll keep an eye on your changes and we can discuss how to handle those as the need arises.
-Allen

tintou:
Answer from Milan:
- giving commit rights to me (via tintou) would help, the done (and coming, if any) changes are strictly to libical-glib
- there are already API changes in the done patches, which makes it harder for the compatibility, but it's unlikely the libical-glib is used too much, thus I'd not be afraid of it
- the offered LIBICAL_GLIB_UNSTABLE_API define mentioned above would probably make things clearer, and I'm in favour to just do it (as an additional change)

tintou:
@winterz thanks for the commit access :tada: 
Two questions that Milan would like you to give an explicit go/no-go:
 * The patches for libical-glib are going to break its API/ABI stability. Giving that there are currently no known user of libical-glib, would it be alright to do it. It would also be greatly appreciated to have it in the 3.0 release.
 * Would a `LIBICAL_GLIB_UNSTABLE_API` define guard for libical-glib be alright to prevent people to think that libical-glib is stable for now.

winterz:
> @winterz thanks for the commit access tada
> Two questions that Milan would like you to give an explicit go/no-go:
> 
>     * The patches for libical-glib are going to break its API/ABI stability. Giving that there are currently no known user of libical-glib, would it be alright to do it. It would also be greatly appreciated to have it in the 3.0 release.
> 
>     * Would a `LIBICAL_GLIB_UNSTABLE_API` define guard for libical-glib be alright to prevent people to think that libical-glib is stable for now.

Ok. 
Please mention it in the ReadMe that we currently do not guarantee source or binary compatibility in the libical-glib

PR #374, opened 2019-04-07
Add "MSEC" recurrence rule type for millisecond precision intervals

kirgene:
Usage example: "FREQ=MSEC;INTERVAL=250" - repeat event each 250ms.

### DIFF
--- a/src/libical-glib/api/i-cal-recurrence.xml
+++ b/src/libical-glib/api/i-cal-recurrence.xml
@@ -15,6 +15,7 @@
 -->
 <structure namespace="ICal" name="Recurrence" native="struct icalrecurrencetype" is_bare="true" default_native="i_cal_recurrence_new_default ()">
      <enum name="ICalRecurrenceFrequency" native_name="icalrecurrencetype_frequency" default_native="I_CAL_NO_RECURRENCE">
+        <element name="ICAL_MSEC_RECURRENCE"/>
         <element name="ICAL_SECONDLY_RECURRENCE"/>
         <element name="ICAL_MINUTELY_RECURRENCE"/>
         <element name="ICAL_HOURLY_RECURRENCE"/>
--- a/src/libical-glib/api/i-cal-time.xml
+++ b/src/libical-glib/api/i-cal-time.xml
@@ -164,6 +164,7 @@
         <parameter type="const gint" name="hours" comment="difference of hours adjusted"/>
         <parameter type="const gint" name="minutes" comment="difference of minutes adjusted"/>
         <parameter type="const gint" name="seconds" comment="difference of seconds adjusted"/>
+        <parameter type="const gint" name="msecs" comment="difference of milliseconds adjusted"/>
         <comment xml:space="preserve">like icaltime_compare_tz, but only use the date parts.</comment>
     </method>
     <method name="i_cal_time_normalize" corresponds="icaltime_normalize" since="1.0">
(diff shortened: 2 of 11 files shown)


winterz:
is MSEC frequency part of an RFC?  or is this something you need for your own project?
I'm not against adding such a feature, but would like to know why we should support it.

also I'd like to see feature specific tests added to regression.c

and fix the crashes, of course

ksmurchison:
I'm not aware of any work being done to extend ICalendar to support milliseconds in either the IETF or CalConnect.  

I have concerns with backwards compatibility with this extension.  How does a generator of a VEVENT with DTSTART using milliseconds know that a consumer can parse and understand the resource?  What is a consumer that is unaware or doesn't understand milliseconds supposed to do with such a resource?  I'd almost be inclined to define a new VALUE parameter type for time with milliseconds.

ksmurchison:
Regarding backwards compatibility, rather than extending the DATE-TIME value syntax, we might consider using the new FRACSEC parameter, which will be defined in the upcoming 
draft-ietf-calext-jscalendar-icalendar-01.  Note that I am attempting to have the name changed to SUBSECOND.

2.1.  FRACSEC parameter

   Parameter name:  FRACSEC

   Purpose:  This parameter is used to define fractional seconds for
      time values and durations.  FRACSEC MUST NOT be used in date-time
      calculations or comparisons in  iCalendar.  It is meant to
      preserve time precision on time values and duration with sub-
      second precision, without increasing the time value range within
      iCalendar.

   Description:  This parameter MAY be specified on properties of type
      DATE-TIME or DURATION.  The integral part of the float value MUST
      be zero.  The value MUST NOT be negative.  iCalendar
      implementations SHOULD ignore this parameter in date time
      arithmetic.  Implementations MUST ignore presence of the FRACSEC
      parameter on RECURRENCE-ID properties when determining recurrence
      overrides.  If present on a RECURRENCE-ID property, its value MUST
      match the FRACSEC parameter value on the DATE-TIME property that
      defines the reference point for the recurring instances.

   Format Definition:

   This parameter is defined by the following notation:

                                       fracsec-param = float

   Example:  DTSTART;FRACSEC=0.03:20190605T133015


So, your example would then look like the following:

BEGIN:VCALENDAR
BEGIN:VEVENT
DTSTART;FRACSEC=0.250:20190408T123000
DURATION;FRACSEC=0.050:PT0S
RRULE:FREQ=MSEC;INTERVAL=250;COUNT=10
END:VEVENT
END:VCALENDAR


Unfortunately, there is no way to extend RRULE without breaking backwards compatibility, as we have found with the RSCALE extension (RFC 7529).  So  a new MSECONDLY or MILLISECONDLY frequency is fine by me.  We might also need a way to extend the UNTIL part to account for fractional seconds, perhaps SUBSECOND-UNTIL or some such.

ksmurchison:
Do we think we need more precision than milliseconds (e.g, micro-, nano-, pico-)?

kirgene:
I think precision higher than a millisecond is overkill, at least for now.
Anyway, this PR gives a possibility to add such a feature later: `time_t` type is replaced with `timespec_t` to provide needed precision.

winterz:
@kirgene the patch isn't applying cleanly to master for me.
I wanted test it locally first.
can you update the patch?

kirgene:
@winterz I've updated the PR

ksmurchison:
FYI, there is an effort to write a new RFC specifying subsecond support in iCalendar.  I believe that the spec will be more generic than just milliseconds.  Therefore, I think we should hold off on merging this PR right now until we see some text for the spec.

winterz:
ok then.  sorry @kirgene 
I was excited to finally get this in.

PR #375, opened 2019-04-09
Fix memory leak when parsing broken file

tsdgeos:


ksuther:
I believe #382 is a slightly better approach to fixing these leaks (and it'll log a warning when the situation occurs).

tsdgeos:
Probably, i spent 5 minutes from knowing ical code, so my fix is more a hot-fix than a fix. Will leave to actual maintainers to decide which fix is better, i don't really care if this is discarded.

winterz:
proper fix committed for #382 
hopefully oss-fuzz is happy

PR #381, opened 2019-05-11
Set a maximum length per line before calling icalparser_add_line

ksuther:
This provides a hard limit on parsing to prevent unbounded memory usage or hanging when expanding parameters or properties.
RFC 5545 says lines should not exceed 75 octets, so this still provides a pretty large buffer (8192 bytes) for files to be non-conformant and still parse successfully.
oss-fuzz issue 14180, 14171, and 14591.

There are alternate ways that we could allow successfully parsing crazy situations with large numbers of parameters and properties (such as referencing counting parameters to avoid unnecessarily duplicating icalparameter structs), but this seems like a fairly reasonable solution. The only drawback that it would fail to parse data that doesn't conform to the SHOULD NOT in the spec, and only by a very large margin.

### DIFF
--- a/src/libical/icalparser.c
+++ b/src/libical/icalparser.c
@@ -45,6 +45,7 @@
 #include <stdlib.h>
 
 #define TMP_BUF_SIZE 80
+#define MAX_LINE_LENGTH 8192 /* the maximum number of chars per parser line */
 
 struct icalparser_impl
 {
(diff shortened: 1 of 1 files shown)


src/libical/icalparser.c
```diff
@@ -641,7 +641,15 @@ icalcomponent *icalparser_parse(icalparser *parser,
     do {
         line = icalparser_get_line(parser, line_gen_func);
 
-        if ((c = icalparser_add_line(parser, line)) != 0) {
+        if (line != 0 && strnlen(line, 8192) >= 8192) {
```
winterz: define MAXLINELEN 8192

ksuther: Done, thanks.

winterz:
committed in 043e5e05821cfa97e160c36e740667fa256a48ea
in both the master and 3.0 branches
will be included with the 3.0.5 release

ksuther:
@winterz After thinking about this again I didn't fix this correctly. I'm checking the length of the line after unfolding it rather than before unfolding, which will break things. This also means the fuzzer could potentially learn to fold parameters and multiple values across multiple lines and still run into the same problem. Can you revert this and I'll think up some other way of handling this? Sorry about that!

ksuther:
After poking at this some more I'm not thrilled with the approach of limiting input by length, as there could be really long valid inputs that are folded across multiple lines (such as an ATTACH property with a large inline file).

The core issue is that the number of parameters on a value can grow rapidly when there's many parameters on multiple values. For example, a property could have 1000 parameters, and then that property could contain 500 values. That means there's 500 * 1000 parameters being generated during parsing because each value gets a clone of the 1000 parameters (they're all identical, but they get duplicated in memory).

We could limit the number of parameters per property or the number of multiple values per property to limit memory usage. Another fix is to reference count parameters so we only need 1000 parameters that the multiple values can each reference, but that's a ton of work for a very specific situation (especially in C).

Any suggestions?

ksmurchison:
I'd say pick reasonable limits on the number of parameters and values per property.  1000 and 500 respectively seems excessive.  20 and 100 seems reasonable to me without being too restrictive.

PR #382, opened 2019-05-11
Fixed leaks when encountering END before BEGIN

ksuther:
Fixes memory leaks caused by the parser behaving incorrectly when the level is negative.
oss-fuzz issue 14480, 14151, 14152, 14153, 14155.

It'll print a warning and then continue parsing. Any lines before a BEGIN will get ignored, so this is pretty lenient and will still parse anything after a valid BEGIN (this is the current behavior as well). No change in behavior, just fixes the leaks.


winterz:
committed in 56621f1ebd351226c8bd41688a46b24d405c5889
for the 3.0 and master branches
will be part of the 3.0.5 release

PR #384, opened 2019-05-13
Cap number of allowed parameters and properties

ksuther:
Reverted the changes from #381 and trying a different approach.

I set the max parameters to 100 and max multiple values per property to 500. That changed the peak memory usage from 2.4 GB to 47 MB (20 parameters and 100 values as suggested by @ksmurchison was even less at 8 MB — at least for the file I was testing — but 50 MB still seems reasonable for intentionally malformed files).


winterz:
committed in master and 3.0 branches

PR #385, opened 2019-05-13
Fix a use after free

### LINKED ISSUE #383: Fix for coverity exposed memory leak
562cce0fdc17fd68f2dd7711f2b737ecadf00892 exposed oss-fuzz issue 14156. Reverting it fixes the leak and seems reasonable to me, but I don't know what the original warning was for.

ksuther:
Found the problem related to 562cce0 and why it was added in the first place. Fixes #383.


winterz:
committed in master and 3.0 branch with sha 584a7b4832551d75e4dcacc7cfece7ddd0fdc74c

PR #389, opened 2019-05-17
Include invalid parameters in the maximum number of allowed parameters

ksuther:
Fixes hangs when iterating lots of invalid parameters (valid parameters already have a cap from f6c64896988644a7c6d4d5060f6ab574c2a62fbf)
oss-fuzz issue 14809


winterz:
committed sha 87cd4152275e4c9aa37df372c00b94cc1992e04d
both master and 3.0

Thanks Kent

PR #395, opened 2019-06-06
In parser_get_next_char, check first character

darktrojan:
Checking `prev_char` is not `\0` immediately after making it `\0` means the block is never run for the first character in `str`, thus `quote_mode` might be wrong or the return value might be wrong.

### DIFF
--- a/src/libical/icalparser.c
+++ b/src/libical/icalparser.c
@@ -152,7 +152,7 @@ static char *parser_get_next_char(char c, char *str, int qm)
     char prev_char = 0;
 
     while (next_char != '\0') {
-        if ((prev_char != '\0') && (prev_char != '\\')) {
+        if (prev_char != '\\') {
             if (qm == 1 && next_char == '"') {
                 /* Encountered a quote, toggle quote mode */
                 quote_mode = !quote_mode;


ksmurchison:
I think your fix is correct, but have you run into (or can you manufacture) data that gets parsed incorrectly without your fix?  I'd like to add a test for this.

ksmurchison:
Actually, making this change breaks existing tests, so without a test case where the current code fails to correctly parse, I'm reluctant to make a change.

darktrojan:
This change was made to stop libical failing [this test](https://searchfox.org/comm-central/rev/9f64b3c10baf73678700dbb7b82020ce0f54ae34/calendar/test/unit/test_bug1209399.js#58) (specifically "Test 4") in Thunderbird. We're in the middle of a major release right now, so I haven't got time to take it further.

ksmurchison:
I don't think Test 4 is legal per RFC5545.  It doesn't appear to me that you can nest DQUOTE:

     param-value   = paramtext / quoted-string

     paramtext     = *SAFE-CHAR

     quoted-string = DQUOTE *QSAFE-CHAR DQUOTE

     QSAFE-CHAR    = WSP / %x21 / %x23-7E / NON-US-ASCII
     ; Any character except CONTROL and DQUOTE

     SAFE-CHAR     = WSP / %x21 / %x23-2B / %x2D-39 / %x3C-7E
                   / NON-US-ASCII
     ; Any character except CONTROL, DQUOTE, ";", ":", ","

ksmurchison:
This commit may be what you're looking for: https://github.com/libical/libical/commit/34cd999317d7a14b03108a112ad1d732db81a202
It allows for empty parameter values.  
Every time I tried to mess with parser_get_next_char() tests failed and/or spun.

winterz:
closing this one as we cannot accept the request.

PR #396, opened 2019-06-14
Fix leak found by fuzzer (oss-fuzz issue 15195)

ksuther:
Caused when multiple RSCALE parameters are encountered (which isn't expected, but possible)


ksmurchison:
I wonder if we shouldn't just return -1 rather than accepting multiple RSCALEs

ksuther:
That would fix it too, I defer to you on that one :)

ksmurchison:
I'll do that after I figure out which of my recent libical commits is causing my CalDAV server to crash during some regression tests.

PR #398, opened 2019-06-26
Remove double #include’d headers

dilyanpalauzov:
… as found by deheader.  Some headers are included in generated .c files twice.

### DIFF
--- a/src/libicalvcal/vcc.c
+++ b/src/libicalvcal/vcc.c
@@ -216,9 +216,6 @@ DFARS 252.227-7013 or 48 CFR 52.227-19, as applicable.
 #include <config.h>
 #endif
 
-#include <string.h>
-#include <stdio.h>
-#include <stdlib.h>
 #include <ctype.h>
 #include "vcc.h"
 
--- a/src/libicalvcal/vcc.y
+++ b/src/libicalvcal/vcc.y
@@ -117,9 +117,6 @@ DFARS 252.227-7013 or 48 CFR 52.227-19, as applicable.
 #include <config.h>
 #endif
 
-#include <string.h>
-#include <stdio.h>
-#include <stdlib.h>
 #include <ctype.h>
 #include "vcc.h"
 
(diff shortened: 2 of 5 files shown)


src/libicalvcal/vcc.c
```diff
@@ -216,9 +216,7 @@ DFARS 252.227-7013 or 48 CFR 52.227-19, as applicable.
 #include <config.h>
 #endif
 
-#include <string.h>
 #include <stdio.h>
-#include <stdlib.h>
```
winterz: I *think* removing the #include <string.h> and #include <stdlib.h> starting at line 1403 is a better fix.  can you try that?

src/libicalvcal/vcc.y
```diff
@@ -117,9 +117,7 @@ DFARS 252.227-7013 or 48 CFR 52.227-19, as applicable.
 #include <config.h>
 #endif
 
-#include <string.h>
 #include <stdio.h>
-#include <stdlib.h>
```
winterz: I wonder if you need stdio.h either.  if you include vcc.h first.  in fact, I would like the include order to be (if possible):
config.h
vcc.h
<standard C headers>

src/libicalvcal/vcc.c
dilyanpalauzov: No.  The lines are removed from vcc.y, and then yacc/bison generates vcc.c .  If I remove the lines 1403…1404, the change will appear again (moving #include <stdlib.h> from line 1403 to line 221).

src/libicalvcal/vcc.y
dilyanpalauzov: Without `#include <stdio.h>` *vcc.c* does compile. I have included also that change.

src/libicalvcal/vcc.c
winterz: ok

src/libicalvcal/vcc.y
winterz: great. thanks

PR #403, opened 2019-08-02
Add option to not build tests

nicolasfella:
I tried to build libical for Android and succeeded, except for a build failure in the tests. While I obviously could try to fix the issue I thought that not building the tests might be desirable in other situations as well as many projects have a similar option


nicolasfella:
The Android build error is not trivial to fix as getpwent() is only available in Android 26+

winterz:
Hi Nico.
a couple things:
- yes, I like the BUILD_TESTING option
 - rename from BUILD_TESTING to LIBICAL_BUILD_TESTING
- please document at the top of CMakeLists.txt (like I do for the other options)
- for getpwent, how about we don't build the stow test on Android?  if(NOT ANDROID) (or whatever CMake uses for Android builds)

winterz:
the reason I want to use "LIBICAL_BUILD_TESTING" is because I've been wanting to namespace the CMake options with the "LIBICAL_" prefix.

nicolasfella:
I don't think I have permission to merge this

Issue #428, opened 2020-05-01
RRULE with BYYEARDAY and BYMONTHDAY expands differently from other implementations


ksuther:
A user encountered this RRULE in the wild:

`FREQ=MONTHLY;BYYEARDAY=-1;BYMONTHDAY=-1`

According to RFC 5545 it's invalid:

> The BYYEARDAY rule part MUST NOT be specified when the FREQ rule part is set to DAILY, WEEKLY, or MONTHLY.

However, Google Calendar and Apple Calendar both happily ignore the `BYYEARDAY` and treat it as the last day of every month. libical seems to treat it as invalid and doesn't return any recurrences.

Any opinions on changing the behavior to improve interoperability?

ksmurchison:
Do we happen to have the PRODID for the generator of this garbage?  Just curious.

ksuther:
Came from Google Calendar, but Google Calendar rewrites all events to its PRODID so I don't know where it originally came from. I'll update this if I'm able to find out the source.

ksmurchison:
The PR above simply ignores BYYEARDAY if used with FREQ=MONTHLY, regardless of whether BYMONTHDAY is specified and fixes the subject issue.
The question is should FREQ=MONTHLY;BYYEARDAY=-1 still be an error?  With this PR it won't be and BYYEARDAY will be ignored.
The bigger question is should we extend the ignore behavior to all BYxxx parts that are illegal? (BYWEEKNO used with anything other than YEARLY, BYYEARDAY used with DAILY or WEEKLY, BYMONTHDAY used with WEEKLY)

ksmurchison:
Actually, I see that I had already created a PR for this issue.  The questions above still stand

ksmurchison:
New API to allow the application to decide how to handle invalid RRULEs in PR #445

ksuther:
> The bigger question is should we extend the ignore behavior to all BYxxx parts that are illegal? (BYWEEKNO used with anything other than YEARLY, BYYEARDAY used with DAILY or WEEKLY, BYMONTHDAY used with WEEKLY)

Left this as a comment in #445, but duplicating it here so the context is kept together:

I haven't thought through every permutation, but since this is working around a very specific problem I'm inclined to say we have a specific set of workarounds to allow interoperability but otherwise leave things strict. That way we have exact tests and documentation for why we accept particular incorrect behavior while not inadvertently introducing our own (and other parsers then having to account for libical's quirky behavior in the future).

PR #429, opened 2020-05-01
issue #428: ignore BYYEARDAY when used with FREQ=MONTHLY

ksmurchison:
Possible fix if we choose to support this


ksmurchison:
Are there any other combinations that we want to ignore, e.g. FREQ=DAILY;BYYEARDAY  ?

ksmurchison:
@winterz @ksuther  Before I merge, any thoughts on my question above re: other bogus combos to ignore?

ksuther:
We could mark all of the N/A rules as ignored. This would mean ignoring BYWEEKNO and BYMONTHDAY and adding BYYEARDAY in a couple more cases.

I don't know if that would get us into trouble with valid recurrences somehow, but it would make RRULE evaluation very lenient.

I can't think of any obvious problem with doing this, as just pretending they don't exist doesn't affect the valid parts of the RRULE.

```
      The term "N/A" means that the corresponding BYxxx rule part MUST
      NOT be used with the corresponding FREQ value.

      BYDAY has some special behavior depending on the FREQ value and
      this is described in separate notes below the table.

   +----------+--------+--------+-------+-------+------+-------+------+
   |          |SECONDLY|MINUTELY|HOURLY |DAILY  |WEEKLY|MONTHLY|YEARLY|
   +----------+--------+--------+-------+-------+------+-------+------+
   |BYMONTH   |Limit   |Limit   |Limit  |Limit  |Limit |Limit  |Expand|
   +----------+--------+--------+-------+-------+------+-------+------+
   |BYWEEKNO  |N/A     |N/A     |N/A    |N/A    |N/A   |N/A    |Expand|
   +----------+--------+--------+-------+-------+------+-------+------+
   |BYYEARDAY |Limit   |Limit   |Limit  |N/A    |N/A   |N/A    |Expand|
   +----------+--------+--------+-------+-------+------+-------+------+
   |BYMONTHDAY|Limit   |Limit   |Limit  |Limit  |N/A   |Expand |Expand|
   +----------+--------+--------+-------+-------+------+-------+------+
   |BYDAY     |Limit   |Limit   |Limit  |Limit  |Expand|Note 1 |Note 2|
   +----------+--------+--------+-------+-------+------+-------+------+
   |BYHOUR    |Limit   |Limit   |Limit  |Expand |Expand|Expand |Expand|
   +----------+--------+--------+-------+-------+------+-------+------+
   |BYMINUTE  |Limit   |Limit   |Expand |Expand |Expand|Expand |Expand|
   +----------+--------+--------+-------+-------+------+-------+------+
   |BYSECOND  |Limit   |Expand  |Expand |Expand |Expand|Expand |Expand|
   +----------+--------+--------+-------+-------+------+-------+------+
   |BYSETPOS  |Limit   |Limit   |Limit  |Limit  |Limit |Limit  |Limit |
   +----------+--------+--------+-------+-------+------+-------+------+
```

winterz:
heck, I'm still dealing with the idea that there are lots of calendars out there with missing UIDs on their components.

I'm happy with whatever you and @ksuther think is best.

ksmurchison:
I hate to blindly overlook bad client behavior without good reason, but I also subscribe to lenient in what we accept.  Have you tried running the test suite with all of the ILLEGAL changed to IGNORE?

minichma:
I feel that there's no good 'fits all' solution. While I see, that there are good reasons for being compatible with Apple and Google, there are certainly scenarios where being more stringent is the right way to go. What about making the behavior (lenient vs stringent) configurable, either at build or in the field? If configured lenient, we would ignore *all* `ILLEGAL` combinations and if configured stringent, we would reject all of them.

ksmurchison:
I had thought about making it configurable at compile time.  I'll take a look at that

ksmurchison:
Superseded by #445

PR #430, opened 2020-05-09
Fix inconsistencies and typos in icaltime docs

mcclurgm:
There are many inconsistencies in the documentation of icaltime that I have found, and tried to fix. Some are between the main `icaltime.c` and `icaltime.h` files, which led to duplicate sentences in the main documentation. Others are with bindings, in particular `libical-glib`. When there were conflicts, I tried to pick the most descriptive, and adhere to a consistent style with the rest of the documentation.

### DIFF
--- a/src/libical-glib/api/i-cal-time.xml
+++ b/src/libical-glib/api/i-cal-time.xml
@@ -48,7 +48,7 @@
     </method>
     <method name="i_cal_time_new_today" corresponds="icaltime_today" kind="others" since="1.0">
         <returns type="ICalTime *" annotation="transfer full" comment="The newly created #ICalTime" />
-        <comment xml:space="preserve">Create a #ICalTime representing today</comment>
+        <comment xml:space="preserve">Create a #ICalTime representing today, with #ICalTime.is_date set.</comment>
     </method>
     <method name="i_cal_time_new_from_timet_with_zone" corresponds="icaltime_from_timet_with_zone" kind="others" since="1.0">
         <parameter type="const time_t" name="v" comment="The seconds past since epoch time"/>
--- a/src/libical-glib/api/i-cal-time.xml
+++ b/src/libical-glib/api/i-cal-time.xml
@@ -158,7 +158,7 @@
         <parameter type="const ICalTime *" name="b" annotation="in, transfer none" comment="The #ICalTime to be compared"/>
 	<parameter type="ICalTimezone *" name="zone" annotation="nullable" comment="The target timezone"/>
         <returns type="gint" comment="-1, 0, or 1 to indicate that a less than b, a==b or a larger than b." />
-        <comment xml:space="preserve">Like icaltime_compare, but only use the date parts; accepts timezone.</comment>
+        <comment xml:space="preserve">Like i_cal_time_compare(), but only use the date parts; accepts timezone.</comment>
     </method>
     <method name="i_cal_time_adjust" corresponds="icaltime_adjust" since="1.0">
         <parameter type="ICalTime *" name="tt" native_op="POINTER" comment="The #ICalTime to be set"/>
--- a/src/libical-glib/api/i-cal-time.xml
+++ b/src/libical-glib/api/i-cal-time.xml
@@ -166,7 +166,7 @@
         <parameter type="const gint" name="hours" comment="difference of hours adjusted"/>
         <parameter type="const gint" name="minutes" comment="difference of minutes adjusted"/>
         <parameter type="const gint" name="seconds" comment="difference of seconds adjusted"/>
-        <comment xml:space="preserve">Add or subtract a number of days, hours, minutes and seconds from an icaltimetype.</comment>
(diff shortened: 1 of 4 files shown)


src/libical-glib/api/i-cal-time.xml
```diff
@@ -143,20 +145,20 @@
...
-        <comment xml:space="preserve">like i_cal_time_compare, but only use the date parts.</comment>
+        <comment xml:space="preserve">Like i_cal_time_compare, but only use the date parts.</comment>
     </method>
     <method name="i_cal_time_compare_date_only_tz" corresponds="icaltime_compare_date_only_tz" since="1.0">
         <parameter type="const ICalTime *" name="a" annotation="in, transfer none" comment="The #ICalTime to be compared"/>
         <parameter type="const ICalTime *" name="b" annotation="in, transfer none" comment="The #ICalTime to be compared"/>
 	<parameter type="ICalTimezone *" name="zone" annotation="nullable" comment="The target timezone"/>
         <returns type="gint" comment="-1, 0, or 1 to indicate that a less than b, a==b or a larger than b." />
-        <comment xml:space="preserve">like i_cal_time_compare_tz, but only use the date parts.</comment>
+        <comment xml:space="preserve">Like icaltime_compare, but only use the date parts; accepts timezone.</comment>
```
tintou, line 161:
```suggestion
        <comment xml:space="preserve">Like i_cal_time_compare_tz(), but only use the date parts; accepts timezone.</comment>
```
Please keep it with the libical-glib namespace :slightly_smiling_face:

mcclurgm: Good catch, thanks! I don't know of a function called `i_cal_time_compare_tz` though, do you mean `i_cal_time_compare`? I know it was in the earlier version, but afaik it doesn't exist, maybe it's a deprecation? Also, I don't see the use of parens anywhere else, so to be consistent with the rest I think that it should not include that. Does this look better?

```suggestion
        <comment xml:space="preserve">Like i_cal_time_compare, but only use the date parts; accepts timezone.</comment>
```

tintou: Indeed `i_cal_time_compare_tz` doesn't exist so it is probably indeed `i_cal_time_compare`. All the reference to functions in libical-glib should be using foo_bar() as this documentation is built using GTK-Doc and thus should be following the [supported syntax](https://developer.gnome.org/gtk-doc-manual/stable/documenting_syntax.html). In this case, having a foo_bar() will allow GTK-Doc to automatically generate a link to the right function.

src/libical-glib/api/i-cal-time.xml
```diff
@@ -143,28 +145,28 @@
...
     </method>
     <method name="i_cal_time_adjust" corresponds="icaltime_adjust" since="1.0">
         <parameter type="ICalTime *" name="tt" native_op="POINTER" comment="The #ICalTime to be set"/>
         <parameter type="const gint" name="days" comment="difference of days adjusted"/>
         <parameter type="const gint" name="hours" comment="difference of hours adjusted"/>
         <parameter type="const gint" name="minutes" comment="difference of minutes adjusted"/>
         <parameter type="const gint" name="seconds" comment="difference of seconds adjusted"/>
-        <comment xml:space="preserve">like i_cal_time_compare_tz, but only use the date parts.</comment>
+        <comment xml:space="preserve">Add or subtract a number of days, hours, minutes and seconds from the @timetype.</comment>
```
tintou, line 169:
```suggestion
        <comment xml:space="preserve">Add or subtract a number of days, hours, minutes and seconds from @tt.</comment>
```
Or change
```
<parameter type="ICalTime *" name="timetype" native_op="POINTER" comment="The #ICalTime to be set"/>
```

src/libical-glib/api/i-cal-time.xml
mcclurgm: Should I change the other function references to include the `()` token then too? Thanks for helping me with this, I'm new to GTK-Doc.

(A suggestion so I can stay within the GitHub system)
```suggestion
        <comment xml:space="preserve">Like i_cal_time_compare(), but only use the date parts; accepts timezone.</comment>
```

mcclurgm:
Sorry about the rush of new commits, once you pointed those out I realized more issues and couldn't let it go! :slightly_smiling_face: I think this should be everything now.

tintou:
Thanks for this :slightly_smiling_face:

winterz:
I'd wonder if you could make a similar patch against the 3.0 branch?
I tried your patch against 3.0 and got some conflicts.

Really appreciate any improvements to the API documentation.
Thank you.

mcclurgm:
You're welcome! To be clear, do you mean the separate 3.0 branch? I'm brand new to this project, so I don't know how it's laid out.

winterz:
yes, there's a 3.0 branch already.

winterz:
@mcclurgm I'm closing this one.  your fixes are in the 3.0 branch.  I will merge them forward into the master branch.

PR #431, opened 2020-06-30
Consistency, grammar, typo fixes to 3.0 docs

mcclurgm:
This is a version of #430 for the 3.0 branch. This is purely grammar fixes and the like, and doesn't include some of the other content changes from that PR. I intend to create another PR for that separately, so I can have smaller diffs.

If any of the style choices (verb form, punctuation) doesn't follow the style of this project, let me know and I'll fix that. I didn't see anything to indicate a particular style, so I followed what seem to be standard practices.

PR #435, opened 2020-07-27
Fix conflicts, new style issues from #431 merge

mcclurgm:
Double checking that the merge didn't create any issues. I looked for differences between the end state of #431 branch and master, and fixed any relevant places that were caused by the merge or that were new and needed similar fixing.

I have a couple questions to be sure everything's in good shape, then good to go.

- [x] There's conflicting information in `icalrecurrencetype_day_day_of_week` comment. I called it out in the diff, so could someone more familiar double check which one is correct?
- [x] In the branch, I fixed some grammar errors in the test data. I should have asked before doing so, but is that something I can do and won't mess up the tests?
- [x] In `astime.h`, I'm not sure if the doc comments for the functions there are still correct, since they were made for what look like completely different functions. Here's their current state:
```c
/** @brief Converts Julian date to year,mo,da. */
void caldat_int(UTinstantIntPtr);

/** @brief Sets Julian day from year,mo,da. */
void juldat_int(UTinstantIntPtr);
```


winterz:
> Double checking that the merge didn't create any issues. I looked for differences between the end state of #431 branch and master, and fixed any relevant places that were caused by the merge or that were new and needed similar fixing.
> 
> I have a couple questions to be sure everything's in good shape, then good to go.
> 
>     * [ ]  There's conflicting information in `icalrecurrencetype_day_day_of_week` comment. I called it out in the diff, so could someone more familiar double check which one is correct?

trash the documentation for icalrecurrencetype_day_day_of_week in icalrecur.c and stick with its doc found in icalrecur.h

> 
>     * [ ]  In the branch, I fixed some grammar errors in the test data. I should have asked before doing so, but is that something I can do and won't mess up the tests?

I didn't notice any changes that affected the test data.
As long as you change only comments I don't think there will be a problem.

> 
>     * [ ]  In `astime.h`, I'm not sure if the doc comments for the functions there are still correct, since they were made for what look like completely different functions. Here's their current state:
> 
> 
> ```c
> /** @brief Converts Julian date to year,mo,da. */
> void caldat_int(UTinstantIntPtr);
> 
> /** @brief Sets Julian day from year,mo,da. */
> void juldat_int(UTinstantIntPtr);
> ```
right. you can move the docs for both those from caldate.c into their rightful place in astime.h
(don't ask me why caldate.c isn't called astime.c or somesuch.  before my time.)

mcclurgm:
Just pushed a commit fixing most of these. As for the test:

> I didn't notice any changes that affected the test data.
As long as you change only comments I don't think there will be a problem.

I might not have been clear, this was from the original PR. Here's an example: https://github.com/libical/libical/pull/431/files#diff-6e96281dd4261bac04cc360f85089f80. This was from a test data file, which I imagine could mess things up if not handled properly (assuming the comment is used as part of a test). I'm not familiar with the testing setup in this project though.

winterz:
> Just pushed a commit fixing most of these. As for the test:
> 
> > I didn't notice any changes that affected the test data.
> > As long as you change only comments I don't think there will be a problem.
> 
> I might not have been clear, this was from the original PR. Here's an example: https://github.com/libical/libical/pull/431/files#diff-6e96281dd4261bac04cc360f85089f80. This was from a test data file, which I imagine could mess things up if not handled properly (assuming the comment is used as part of a test). I'm not familiar with the testing setup in this project though.

Aha.  Right.  Best restore src/Net-ICal-Libical/test-data

mcclurgm:
Done. 

This is all I was able to find, so I don't think there are any significant issues left from the merge. All the differences between master and the branch from the PR look like genuine changes that came from someone else. And thanks for your help with fixing the duplicate signatures in the glib files.

winterz:
Thanks @mcclurgm 
At the moment I only publish the documentation for the 3.0 branch on libical.io.  I think that's correct as that's the version almost everyone should be using.

Issue #394, opened 2019-06-04
Cannot enable glib library if cross compiling


rossburton:
The `src-generator` binary is built with the target compiler so can't be run during the build:

```
| /bin/sh: 1: [...]/build/bin/src-generator: Exec format error
```

```
$ file bin/src-generator
bin/src-generator: ELF 32-bit LSB shared object, ARM, EABI5 version 1 (SYSV), dynamically linked, interpreter /lib/ld-linux-armhf.so.3, for GNU/Linux 3.2.0, not stripped
$ uname -a
Linux flashheart 4.19.0-0.bpo.4-amd64 #1 SMP Debian 4.19.28-2~bpo9+1 (2019-03-27) x86_64 GNU/Linux
```

That binary should be built using the host compiler, not the target compiler.

winterz:
thanks for the report.
for now you should be able to pass -DICAL_GLIB=False to cmake .  I doubt you need the libical-glib library.

winterz:
We just committed some fixes for building with gtk.  could you test?  both the master branch or the 3.0 branch have the fix.

winterz:
we released v3.0.6 today.
please test if this is fixed.

rossburton:
Apologies for not testing sooner.

```
| /bin/sh: 1: /data/poky-tmp/master/work/armv7vet2hf-neon-poky-linux-gnueabi/libical/3.0.6-r0/build/bin/src-generator: Exec format error
```

```
$ file src-generator
src-generator: ELF 32-bit LSB shared object, ARM, EABI5 version 1 (SYSV), dynamically linked, interpreter /lib/ld-linux-armhf.so.3, BuildID[sha1]=9f8ee30a1df80a2c0794f049563fb568d1da7f5c, for GNU/Linux 3.2.0, with debug_info, not stripped
```

Still being built with the cross compiler instead of the native compiler.

rossburton:
FWIW still happens with 3.0.7:
```
| /bin/sh: 1: /data/poky-tmp/master/work/armv7vet2hf-neon-poky-linux-gnueabi/libical/3.0.7-r0/build/bin/src-generator: Exec format error
```

`src-generator` was built with the cross compiler, not the host compiler.

Have you considered moving to Meson, which handles this trivially?

triffid:
I just ran into this issue in 3.0.8

winterz:
should be solved in commit 1197d84b63dce179b55a6293cfd6d0523607baf1
in the 3.0 and master branches

[...]

PR #439, opened 2020-09-28
libical-glib: export src-generator from native build for cross-compilation

lopsided98:
When cross-compiling, executables built for the host architecture cannot be run on the build architecture. CMake doesn't provide something like `CC_FOR_BUILD` from autotools, and recommended way to handle this case is the export the required binary from the native build for use by the cross-compiled build. This PR is roughly based on the method described here: https://gitlab.kitware.com/cmake/community/-/wikis/doc/cmake/CrossCompiling#using-executables-in-the-build-created-during-the-build.

To use this PR, build and install a native build of libical, and then pass `-DIMPORT_GLIB_SRC_GENERATOR=/path/to/native/ical/lib/cmake/LibIcal/GlibSrcGenerator.cmake` to the cross-compiled build.

### DIFF
--- a/src/libical-glib/CMakeLists.txt
+++ b/src/libical-glib/CMakeLists.txt
@@ -1,15 +1,26 @@
 add_definitions(-Dlibical_ical_EXPORTS)
 
-# build the src-generator
-add_executable(src-generator
+# build ical-glib-src-generator
+add_executable(ical-glib-src-generator
   tools/generator.c
   tools/generator.h
   tools/xml-parser.c
   tools/xml-parser.h
 )
 
-target_compile_options(src-generator PUBLIC ${GLIB_CFLAGS} ${LIBXML_CFLAGS} -DG_LOG_DOMAIN=\"src-generator\")
-target_link_libraries(src-generator ${GLIB_LIBRARIES} ${LIBXML_LIBRARIES})
+target_compile_options(ical-glib-src-generator PUBLIC ${GLIB_CFLAGS} ${LIBXML_CFLAGS} -DG_LOG_DOMAIN=\"src-generator\")
+target_link_libraries(ical-glib-src-generator ${GLIB_LIBRARIES} ${LIBXML_LIBRARIES})
+
+install(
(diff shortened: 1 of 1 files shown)


winterz:
looks good.
does this solve https://github.com/libical/libical/issues/394?

winterz:
committed in 1197d84b63dce179b55a6293cfd6d0523607baf1 
branches 3.0 and master

Thanks again!

winterz:
I forgot: we should add instructions for this to our wiki.  I started a one here -> https://github.com/libical/libical/wiki/For-the-Users:-Cross-compiling
can you review?

lopsided98:
That looks good except the relevant section of the linked CMake wiki page is "Using executables in the build created during the build". It's annoying that linking to a specific section doesn't actually work.

Also I was thinking about this PR some more and realized that it makes more sense for the src-generator to go in `CMAKE_INSTALL_LIBEXECDIR`, since it is for internal use only and doesn't need to be on `PATH`. I've created a followup PR: #440

winterz:
@lopsided98 I'm getting some pushback on your patch from the people who maintain libical-glib.  Can you address these issues?

- It is installed into the PREFIX. I do not think it makes any sense to install it, it's used only in the build time and no where else, it
was always meant as a private tool to convert XML API descriptions into a source file, ready for compilation. You can drop it from the install() for sure.
- Its current naming is also kinda bad. Calling it "glib-src-generator" implies it's from the glib library/namespace, which is wrong, this is from libical.
-  A path .../libexec/ical/ feels bad too, as everywhere else is used 'libical', not 'ical'.

lopsided98:
Installing it to `PREFIX` was the point of this patch. It makes it much easier to cross-compile `libical-glib`, particularly from the perspective of a distro maintainer. With this patch, we can just point the cross build at the exported target from a native `libical` package and everything works. If the src generator was not installed, we would have to build a mostly useless native libical alongside the cross-compiled one and manually extract the src generator from it.

winterz:
yes, looks good.
could you move it to the 3.0 branch?

PR #440, opened 2020-10-01
libical-glib: install src-generator to CMAKE_INSTALL_LIBEXECDIR

lopsided98:
Follow-up to #439. The src-generator binary is not useful to end-users, so it really shouldn't be on `PATH`. `libexec` is the standard location for this kind of executable.


winterz:
I merged this one.
commit 1197d84b63dce179b55a6293cfd6d0523607baf1 and 88c43c7d796d8dc5db730f39f3eac14f8c23fffa

PR #444, opened 2020-10-15
issue #428: ignore BYYEARDAY when used with FREQ=MONTHLY (to match Ap…

ksmurchison:
…ple and Google behavor)

### DIFF
--- a/src/libical/icalrecur.c
+++ b/src/libical/icalrecur.c
@@ -991,7 +991,7 @@ int icalrecur_iterator_sizeof_byarray(short *byarray)
 
 enum expand_table
 {
-    UNKNOWN = 0,  // can be used to ignore an otherwise illegal BYxxx rule
+    IGNORE = 0,
     CONTRACT = 1,
     EXPAND = 2,
     ILLEGAL = 3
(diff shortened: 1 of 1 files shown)


minichma:
Why would we ignore `BYYEARDAY` in case of `FREQ=MONTHLY` but not in case of `WEEKLY` or `DAILY`. Or in a more general sense, why would we ignore some illegal combinations while we reject others? Doesn't feel right. Might introduce more confusion than it helps.

src/libical/icalrecur.c
```diff
@@ -991,7 +991,7 @@ int icalrecur_iterator_sizeof_byarray(short *byarray)
 
 enum expand_table
 {
-    UNKNOWN = 0,
+    IGNORE = 0,
```
minichma, line 994: `UNKNOWN` and `IGNORE` have quite different meanings, even if the technical effect might be the same. Maybe a distinct value would make sense.

ksmurchison: Yes, I did exactly that in PR #429.  I had forgotten that I wrote that when I did this one, which I'm going to close.  Please address further comments on that PR

PR #445, opened 2020-10-15
issue #428: runtime setting to ignore invalid BYxxx part combos in RR…

ksmurchison:
…ULEs

I wondering if we should check validity in icalrecurrencetype_from_string()?  That's really the only API that we have for creating an icalrecurrencetype, other than populating the struct manually

### DIFF
--- a/ReleaseNotes.txt
+++ b/ReleaseNotes.txt
@@ -22,6 +22,8 @@ Version 3.1.0 (NOT RELEASED YET):
      + icalcomponent_get_x_name
      + icalcomponent_get_component_name
      + icalcomponent_get_component_name_r
+     * ical_set_invalid_rrule_handling_setting
+     * ical_get_invalid_rrule_handling_setting
  * icaltzutil_get_zone_directory() can use the TZDIR environment to find system zoneinfo
  * Deprecated functions:
      + caldat (replaced by internal function icaldat_int())
--- a/src/libical/icalrecur.c
+++ b/src/libical/icalrecur.c
@@ -142,6 +142,13 @@
 #include <stddef.h>     /* For offsetof() macro */
 #include <stdlib.h>
 
+#if defined(HAVE_PTHREAD)
+#include <pthread.h>
+static pthread_mutex_t invalid_rrule_mutex = PTHREAD_MUTEX_INITIALIZER;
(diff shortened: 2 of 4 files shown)


ksuther (commented):
Returning an error in `icalrecurrencetype_from_string` makes sense to me, as it already returns errors in other scenarios. That would make it easier to use the API for validation instead of discovering that a RRULE is invalid only when you try to use it.

This should also have quirks mode tests for `ICAL_RRULE_IGNORE_INVALID`.

In https://github.com/libical/libical/issues/428#issuecomment-709361693 you asked if we should apply this to all illegal rules. I haven't thought through every permutation, but since this is working around a very specific problem I'm inclined to say we have a specific set of workarounds to allow interoperability but otherwise leave things strict. That way we have exact tests and documentation for why we accept particular incorrect behavior while not inadvertently introducing our own (and other parsers then having to account for libical's quirky behavior in the future).

ksmurchison:
I think if we use this PR rather than #429 then the question of whether we want to be strict or lenient is up to the client.  The default mode will be strict and in the case of BYMONTHDAY + BYYEARDAY, the client can choose to be lenient.

I will look into expanding this option to also work at parse time.

ksmurchison:
@ksuther @winterz  Updated to include parse time check (with change to regression.c which happens to include an invalid RRULE)

ksmurchison:
Where do we stand on this one?

ksuther:
This seems reasonable to me. So far the only mismatched case I've seen is the original one I reported, but we may turn this mode on in Fantastical and see if anyone reports anything else unexpected :)

ksmurchison:
Hmm. I'm wondering if we should strip the invalid BY* parts so we don't output bad RRULEs

ksmurchison:
Updated to ignore invalid BY* part data so icalrecurrencetype_as_string() doesn't output it

ksmurchison:
@winterz @ksuther Looking for one final set of eyes to look at this before I merge.

ksuther:
Late to the party on this one, but it still makes sense to me. I mentioned this before, but I'm curious if any weird unintended parsing get encountered in the wild that don't make any sense at all. I guess time will tell :)

PR #446, opened 2020-10-24
restrictions: VTIMEZONE can contain any number of X properties

dilyanpalauzov:


winterz:
Safe to backport to 3.0?
Also can I get a 1-liner for the ReleaseNotes.txt?

dilyanpalauzov:
` * accept VTIMEZONE with more than one X- properties`

I do not use 3.0, but this is rather semi-urgent.  DAVx⁵ 3.3.4, released on 16 Oct, sends to the server VTIMEZONE definitions with 
```
X-LIC_LOCATION:America/New_York
X-PROLEPTIC-TZNAME:LMT
```
(either without METHOD, or it just triggers the restrictions that apply when no METHOD is set) and libical validation says this is invalid.  https://forums.bitfire.at/topic/2400/ says a new DAVx⁵ version will be made, that has just one X- property,

PR #447, opened 2020-10-25
Move some variables to the read-only section

dilyanpalauzov:
so that they do not allocate additional memory in each binary, that uses libical.


winterz:
I merged this.  not sure why github didn't notice.
commit 323ee40aa91913ac8084969f33a5ed9cf133754d

PR #448, opened 2020-10-30
Fix build with icu-68.1

Polynomial-C:
icu-68.1 removed public macro definitions for TRUE and FALSE

See icu upstream commit: https://github.com/unicode-org/icu/commit/a18df7ba2834afc2b577a1d5a7d906868ade6fa9

### DIFF
--- a/src/libical/icalrecur.c
+++ b/src/libical/icalrecur.c
@@ -143,6 +143,7 @@
 #include <stdlib.h>
 
 #if defined(HAVE_LIBICU)
+#include <stdbool.h>
 #include <unicode/ucal.h>
 #include <unicode/ustring.h>
 #else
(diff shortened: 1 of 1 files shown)


winterz:
do we know for sure that all the world's platforms (let's say within the past 15 years) have stdbool.h?

SoapGentoo:
`stdbool.h` is a C99 header. The problem is, ICU already exposes `stdbool.h` in other headers, hence if you're using the latest ICU, you're implicitly relying on C99 support anyways.
GCC has the `bool` macro in `stdbool.h` since GCC 3.0, which was released in 2001. I think it's safe to say that, yes, most platforms from the past 15 years have a proper `stdbool.h`.

winterz:
see commit a3308a23912bba2db654a8c456165c31888cc897

in case some ancient compiler (like MSVC before version 2013) sees the code.
let me know what you think.
-Allen

SoapGentoo:
MSVC always compiles in C++ mode, so doesn't need the macros for `true`/`false`

winterz:
I tried MSVC 2010 and got this without defining false ourselves:

```
C:\projects\libical\libical\src\libical\icalrecur.c(1024) : error C2065: 'false' : undeclared identifier
C:\projects\libical\libical\src\libical\icalrecur.c(1417) : error C2065: 'false' : undeclared identifier
```
I'll leave the code as is , at least for now.

PR #451, opened 2020-11-01
icalset.c:icalset_new() skip memset() calls

dilyanpalauzov:
As the next line changes all bits in the `data` variable, there is no point to zeroing the `data` variable.


dilyanpalauzov:
No, it does not.

PR #453, opened 2020-11-01
icalrecur.c: move wd_map to read-only section

dilyanpalauzov:
at the expense of allocating three bytes for the ICAL_NO_WEEKDAY key, which are never read.

Pro: all programs using libical at the same time will share the memory for wd_map, instead of having their own copy

Contra: the own copy would be three bytes smaller, if only one program uses libical.

### DIFF
--- a/src/libical/icalrecur.c
+++ b/src/libical/icalrecur.c
@@ -270,10 +270,10 @@ const char *icalrecur_skip_to_string(icalrecurrencetype_skip kind)
     return 0;
 }
 
-static struct wd_map
+static const struct wd_map
 {
     icalrecurrencetype_weekday wd;
-    const char *str;
+    const char str[3];
 } wd_map[] = {
     {ICAL_SUNDAY_WEEKDAY, "SU"},
     {ICAL_MONDAY_WEEKDAY, "MO"},
@@ -282,7 +282,7 @@ static struct wd_map
     {ICAL_THURSDAY_WEEKDAY, "TH"},
     {ICAL_FRIDAY_WEEKDAY, "FR"},
     {ICAL_SATURDAY_WEEKDAY, "SA"},
-    {ICAL_NO_WEEKDAY, 0}
+    {ICAL_NO_WEEKDAY, "  "}
 };
 
 const char *icalrecur_weekday_to_string(icalrecurrencetype_weekday kind)


ksmurchison (commented):
Is there any reason to make ICAL_NO_WEEKDAY anything other than the empty string ("")?

dilyanpalauzov:
This does not change anything on the memory usage, that is three bytes for that string.  It also does not change anything on the fact, that its value is never read.  This spelling makes it double-obvious that it is three bytes.  I can make it also the empty string.

dilyanpalauzov:
Perhaps for-looping over wd_map, based on the size of the array calculated using sizeof, can eliminate the last element of wd_map completely.

ksmurchison:
Subsumed by https://github.com/libical/libical/commit/93f952cd6a4ba445b2a3cef7d6b25eeefb5e79df

dilyanpalauzov:
The same principle is applicable to more `maps`.  The read-write maps in libical currently are:
```
$ nm libical.so|grep " d " 
0000000000085620 d builtin_mutex
00000000000855e0 d days.1
00000000000721b8 d __do_global_dtors_aux_fini_array_entry
00000000000721c0 d __dso_handle
0000000000083dc0 d _DYNAMIC
0000000000073980 d enum_map
0000000000085580 d error_state_map
00000000000721b0 d __frame_dummy_init_array_entry
0000000000085660 d freq_map
0000000000084000 d _GLOBAL_OFFSET_TABLE_
0000000000083c80 d icalmime_local_action_map
0000000000078e20 d icalparameter_map
0000000000079460 d icalrestriction_property_records
0000000000085648 d icaltimezone_minimum_expansion_year.1
0000000000085840 d major_content_type_map
0000000000085a00 d mime_headers
00000000000858e0 d minor_content_type_map
0000000000078ae0 d parameter_map
00000000000721e0 d property_map
0000000000085720 d recur_map
0000000000083120 d request_status_map
00000000000856e0 d skip_map
0000000000085a40 d sspm_action_map
0000000000085980 d sspm_encoding_map
00000000000834e0 d string_map
0000000000085820 d s_zoneinfo_search_paths
0000000000085a68 d __TMC_END__
000000000008564c d unknownTokenHandling
```

Issue #438, opened 2020-09-11
Evaluation of the `BYWEEKNO` Property


christophercrouzet:
Hello!

When looking at your suite of tests for the `libical` recurrence rules, this specific case grabbed my attention:

```
RRULE:FREQ=YEARLY;BYWEEKNO=1,2;UNTIL=20170101T000000Z
DTSTART:20130101T000000
INSTANCES:20130101T000000,20130108T000000,20140101T000000,20140108T000000,20150101T000000,20150108T000000,20160108T000000,20160115T000000
```

I might be misunderstanding something here but the RFC 5545 specification seems ambiguous when it comes to treating this specific case so I ended up looking for a less ambiguous specification and stumbled upon the `JSCalendar` one, [section 4.3.2.1](https://datatracker.ietf.org/doc/html/draft-ietf-calext-jscalendar-27#section-4.3.2.1).

From what I gather from this specification, the instances need to match the day of the week of the `DTSTART` date (Tuesday here) to look like this:

```
20130101T000000
20130108T000000
20131231T000000
20140107T000000
...
```

I'm not here to say that one approach is better than the other and/or to ask for any action to be taken. If possible, I'd only be interested to know:
- what led to the decision of implementing this specific case this way in `libical`?
- whether it would make any sense at all to align `libical`'s implementation's with this IETF draft specification?

For full disclosure, I opened a similar issue on their repo, see: https://github.com/CalConnect/PUBLIC_DRAFTS/issues/18

winterz:
I will leave this to the recurrence experts
@ksmurchison or @ksuther  have any comments on this?

ksmurchison:
Yes, I talked to a colleague who used to work on Apple CalendarServer and is an expert on iCalendar and he agrees that for BYWEEKNO, that the day of week should be inherited from DTSTART (as per JSCalendar).  I haven't had a chance to look at fixing the code yet.

PR #455, opened 2020-11-06
icalrecur.c: fix handling of BYWEEKNO with no BYDAY to use the day of…

ksmurchison:
… week from DTSTART rather than day of month

Fix for issue #438, assuming that we agree that this is the proper handling.  FWIW, I asked a former CalConnect colleague (and CalendarServer developer) about this and he agreed that in this case, it makes more sense to use day of week rather than day of month


winterz:
ok by me.
@ksuther any objections?

please also mention the change in the ReleaseNotes.
Oh, and commit to the 3.0 branch

ksuther (approved):
Seems reasonable to me too.

ksmurchison:
@winterz Do you want me to cherry-pick the other BYWEEKNO changes from master to 3.0?  This one only changes what data we inherit from DTSTART, but the others are needed to fix a host of other BYWEEKNO issues.

winterz:
> @winterz Do you want me to cherry-pick the other BYWEEKNO changes from master to 3.0? This one only changes what data we inherit from DTSTART, but the others are needed to fix a host of other BYWEEKNO issues.

yes please cherry-pick the BYWEEKNO changes from master to 3.0
and don't forget the ReleaseNotes:)

Issue #454, opened 2020-11-06
libical not correctly reading data from tzdb-2020d


aguador:
It appears that libical is not reading timezone data correctly there are multiple TZif2 file format headers, at least one of which is empty and crashes. This has been observed with libical's make test internal tests. For more details see https://gitlab.gnome.org/GNOME/evolution-data-server/-/issues/267#note_952898

dilyanpalauzov:
Please provide a timezone file in iCalendar format that causes the problem.

aguador:
I hate to admit it, but I am not sure how to do that. Is there a converter than can grab to tzdb data and convert it? The Madrid tz is causing the failure on my machine.

dilyanpalauzov:
The IANA tzdb is converted to icalendar file by a vzic tool.  llibical consumes iCalendar file.  The consumed file must be on your computer.

aguador:
Here is the file located in /usr/share/timeinfo/Europe that I presume libical uses.

[Madrid.gz](https://github.com/libical/libical/files/5503796/Madrid.gz)

dilyanpalauzov:
The provided file is in TZif2 format.  My understanding is, that the input/output of libical can be only in iCalendar format.  The source code of libical does not mention anywhere tzif/TZif, so this file is not consumed by libical and problem is not caused by libical.

winterz:
for the record, I'm updating libical's builtin zoneinfo data with tzdata2020d today and don't see any crashes in the libical testsuite.  In fact, the tzdata2020d data is has very few changes from the current builtin zoneinfo.

I agree with @dilyanpalauzov . this must be due to the system install TZif2 files .  apparently we should skip those somehow in libical.  anyone feel like making a  patch for that?

dilyanpalauzov:
Olson 2020b, 2020c and 2020d were released almost simultaneously.  The upstream bug https://gitlab.gnome.org/GNOME/evolution-data-server/-/issues/267 says that the problem comes from dropped pacificnew and the [Announcement for Olson 2020c](http://mm.icann.org/pipermail/tz-announce/2020-October/000060.html) says “tarballs now contain an empty file pacificnew”.

While the PR mentions explicitly 2020d, it might mean changes introduced by 2020c.

aguador:
> While the PR mentions explicitly 2020d, it might mean changes introduced by 2020c.

[...]

PR #457, opened 2020-11-11
icaltzutil_fetch_timezone() should read v2/v3 data when available

ksmurchison:
Possible fix for issue #454

### DIFF
--- a/src/libical/icaltz-util.c
+++ b/src/libical/icaltz-util.c
@@ -85,6 +85,9 @@
 //@cond PRIVATE
 typedef struct
 {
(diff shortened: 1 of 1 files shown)


winterz:
code looks fine and so far my testing looks good.
could we extend our test suite with some v2/v3 data?

for now put this in master and I will cherry-pick to 3.0
and add a blurb to ReleaseNotes.

ksmurchison:
Pretty much all data produced by zic over the last few years is v2/v3.  The only change was that zic now writes essentially nothing in the v1 data area

PR #463, opened 2020-12-14
libical-glib: Use the latest GObject code style for classes

tintou:
Requires GLib 2.44.

GLib 2.44 comes with some handy macros. This also allows developers to use `g_autoptr()` with Clang/GCC.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -374,7 +374,7 @@ if(ICAL_GLIB_VAPI)
   endif()
 endif()
 
-set(MIN_GLIB "2.38")
+set(MIN_GLIB "2.44")
 set(MIN_LIBXML "2.7.3")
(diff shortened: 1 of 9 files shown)


winterz:
when I test compile this patch I see lots of warnings like:

/`home/allen/projects/libical/libical/build-test2-scan/src/libical-glib/i-cal-value.c:237:169: warning: passing argument 1 of ‘I_CAL_OBJECT’ discards ‘const’ qualifier from pointer target type [-Wdiscarded-qualifiers]`

tintou:
Indeed, fixed the warnings

winterz:
I'm thinking we could put this in the 3.0 branch as well.
If you agree, then I'll need a new patch to applies cleanly to the 3.0 branch.

Issue #475, opened 2021-01-25
build generator Ninja with static libs fails on Mac


thierryba:
Hello, I tried libical (both 3.0.8 and then 3.0.9)
if I configure with cmake -GNinja -DSTATIC_ONLY=true -DICAL_GLIB=false ../libical-3.0.8 

I'm getting this

CMake Error:
...
  failed with:

   ninja: error: build.ninja:624: multiple rules generate lib/libical_cxx.a [-w dupbuild=err]

It works fine with Unix Makefiles.

winterz:
verified. happens in 3.0.9 and master as well.

winterz:
`cmake -GNinja -DICAL_GLIB=false` works
`cmake -GNinja -DSHARED_ONLY=true -DICAL_GLIB=false` works
`cmake -GNinja -DSTATIC_ONLY=true -DICAL_GLIB=false` fails

krop:
the issue comes from src/libical/CMakeLists.txt.

libical always builds ical_cxx (line 327) but if SHARED_ONLY is false, it also builds ical_cxx-static and tries to rename the produced library to ical_cxx on !MSVC (https://github.com/libical/libical/blob/master/src/libical/CMakeLists.txt#L342)

kuruczgy:
I think #477 fixes exactly this, please take a look if you can.

thierryba:
@kuruczgy 
Sorry but that did not work
the output of CMake is

`thierry@tb libical % cmake -GNinja -DSTATIC_ONLY=true -DICAL_GLIB=false
CMake Warning:
  No source or binary directory provided.  Both will be assumed to be the
  same as the current working directory, but note that this warning will
  become a fatal error in future CMake releases.


-- The C compiler identification is AppleClang 12.0.0.12000032
-- Detecting C compiler ABI info
-- Detecting C compiler ABI info - done
-- Check for working C compiler: /usr/local/bin/cc - skipped
-- Detecting C compile features
-- Detecting C compile features - done
-- The CXX compiler identification is AppleClang 12.0.0.12000032
-- Detecting CXX compiler ABI info
-- Detecting CXX compiler ABI info - done
-- Check for working CXX compiler: /usr/local/bin/clang++ - skipped
-- Detecting CXX compile features
-- Detecting CXX compile features - done
-- Found Perl: /usr/local/bin/perl (found version "5.32.1") 
-- The following ICU libraries were not found:
--   uc (required)
--   i18n (required)
[...]

PR #477, opened 2021-01-31
Fix building *-static libraries when STATIC_ONLY is enabled.

kuruczgy:
When `STATIC_ONLY` is enabled, both the regular libraries, and the ones with the `-static` postfix are built. This PR aims to fix this.


kuruczgy:
fixes #475

PR #482, opened 2021-02-27
Incorrectly placed comma

AvdN:
There is no comma in the example line, so it cannot be part of the property name. Instead it should be placed after the closing double quote.

### DIFF
--- a/doc/UsingLibical.txt
+++ b/doc/UsingLibical.txt
@@ -72,7 +72,7 @@ content line
 
 ORGANIZER;ROLE=CHAIR:MAILTO:mrbig@host.com
 
-The property name is "ORGANIZER," the value of the property is "mrbig@host.com"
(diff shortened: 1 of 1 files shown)


AvdN:
I realised this is probably generated from the .md

winterz:
I'm not so sure.
I'll apply your patch anyway. just in case.

Does anyone recall if we intended to remove UsingLibical.txt in favor of the markdown version?

AvdN:
> I'm not so sure.
> I'll apply your patch anyway. just in case.
> 
> Does anyone recall if we intended to remove UsingLibical.txt in favor of the markdown version?

There is at least one more "swap" that I encountered while reading through the .txt version which lead me to believe that the .txt is derived from the .md by a faulty program. the same date-time-stamps of the .md and .txt also suggest date. 
I didn't search for makefiles/scripts in which this faulty program might be specified though.

Issue #499, opened 2021-06-22
Build issue if path name has a '.xml' pattern


Lionelon:
In one of my project, the path in which I cloned libical is named test.xml/libical
Because of that, there is a build issue: 
[ 57%] Building C object src/libical-glib/CMakeFiles/ical-glib-static.dir/home/osboxes/tmp/test..c/libical/build/src/libical-glib/i-cal-array..c.o
cc: error: <path>/test.c/libical/build/src/libical-glib/i-cal-array..c: No such file or directory
cc: fatal error: no input files
compilation terminated.

Lionelon:
Done and merged in #500

PR #500, opened 2021-06-22
libical-glib: modify xml->[ch] only for files, not folder name

Lionelon:


winterz:
thanks.  can we close #499 ?

PR #503, opened 2021-08-02
Buildsystem fixes

paweljonskim:
Rationale:
1. `PROJECT_` vars instead of `CMAKE_`
Allows to embed this project into other cmake projects. Using `CMAKE_` variables is considered a bad practice.
2. Removing ical-glib-static target
Could not understand why is this target needed, but when building with other tool than make (`ninja`), multiple targets for one of the libs are found, which results in build error. I believe `STATIC`/`SHARED` version of the lib should be decided by build-in CMAKE arguments and vars
3. Resolve GetNumberFormat only on windows
`GetNumberFormat` is not implemented on wine correctly (https://forum.winehq.org/viewtopic.php?t=27809) which results in error when building. That means - if any linux user has installed wine, the build of `libical` will fail. This check fixes that.


winterz:
I'm curious why you closed this

paweljonskim:
I realised those changes were not enough to build `libical` for all targets in my project. Additionally, I cant spend more time on adjusting `libical` buildsystem to modern cmake proper way, so I decided to make some _dirty_ changes and not suitable for PR to mainline `libical` (just commented out some configuration lines from cmake). I wanted to stick to my local branch so I closed this PR which is connected to it.

I still encourage to rework build system to modern cmake and allow other projects to reuse it (also baremetal, not only windows/linux/bsd). Unfortunatelly my changes in this PR are just a tip of the iceberg.

krop:
I tought about it some times ago (replacing the manual include directories. There was a blocker: the gtk doc generator. This thing is horrible to use and failed completely after the revamp.

winterz:
I would be interested in a patch that makes sure libical can be easily as a git submodule for a larger project.
as a first step, anyway.

krop:
> I would be interested in a patch that makes sure libical can be easily as a git submodule for a larger project.
> as a first step, anyway.

That's quite straightforward, replace CMAKE_{SOURCE,BINARY}_DIR with libical_{SOURCE,BINARY}_DIR

krop:
> > I would be interested in a patch that makes sure libical can be easily as a git submodule for a larger project.
> > as a first step, anyway.
> 
> That's quite straightforward, replace `CMAKE_{SOURCE,BINARY}_DIR` with `libical_{SOURCE,BINARY}_DIR`

Now that I think about this, it shouldn't be needed. externalproject_add shouldn't mix the projects

winterz (commented):
I committed this in 485bb36c10bfda5d22068f1c3bf95be3548d0557

PR #507, opened 2021-09-15
Actually use the icaltimezone_set_tzid_prefix() value, even if empty

ksmurchison:
If an application calls icaltimezone_set_tzid_prefix("") and then icaltiicaltimezone_tzid_prefix(), s_ical_tzid_prefix would be reset to BUILTIN_TZID_PREFIX and returned.

### DIFF
--- a/src/libical/icaltimezone.c
+++ b/src/libical/icaltimezone.c
@@ -75,7 +75,7 @@ static struct _compat_tzids {
 };
 
 /* The prefix to be used for tzid's generated from system tzdata */
-static char s_ical_tzid_prefix[BUILTIN_TZID_PREFIX_LEN] = {0};
+static char s_ical_tzid_prefix[BUILTIN_TZID_PREFIX_LEN] = BUILTIN_TZID_PREFIX;
 
 /** This is the filename of the file containing the city names and
     coordinates of all the builtin timezones. */
@@ -202,9 +202,6 @@ static void icaltimezone_changes_unlock(void)
 
 const char *icaltimezone_tzid_prefix(void)
 {
-    if (s_ical_tzid_prefix[0] == '\0') {
-        strncpy(s_ical_tzid_prefix, BUILTIN_TZID_PREFIX, BUILTIN_TZID_PREFIX_LEN-1);
-    }
     return s_ical_tzid_prefix;
 }
 


ksuther:
The change makes sense to me, but did it cause the regression tests to fail?

I've been using this change locally for years to accomplish the same:
```
-static const char *ical_tzid_prefix = "/freeassociation.sourceforge.net/";
+static const char *ical_tzid_prefix = "";
```

winterz:
Your patch is logical to me.  The current API is horrible.

Can you fix the regression test accordingly?

icaltimezone_set_tzid_prefix() should clear `s_ical_tzid_prefix` if `new_prefix` is NULL or '\0' , no (not ignored like it is now)
The return value of icaltimezone_tzid_prefix() should be `BUILTIN_TZID_PREFIX` unless or until `s_ical_tzid_prefix` is changed by a call to icaltimezone_set_tzid_prefix()

winterz:
@ksmurchison what to do with this?  been a few years now

ksmurchison:
I will look at it today

winterz:
I'll look this weekend.  last time I thought about this was 2.5 years ago.

ksmurchison:
No rush. I'm at IETF in Brisbane and won't look at this again until next week when I'm home

winterz:
@ksmurchison @ksuther I modified the patch.  mostly I fixed and added more regression testing.
let me know what you think.

ksmurchison:
This looks good, but  test_builtin_compat_tzid() still fails for me.

winterz:
> This looks good, but test_builtin_compat_tzid() still fails for me.

can you provide details?  the github CI is happy -- but we could be missing a combination of cmake options that could case a failure I suppose

ksmurchison:
> > This looks good, but test_builtin_compat_tzid() still fails for me.
> 
> can you provide details? the github CI is happy -- but we could be missing a combination of cmake options that could case a failure I suppose

@winterz What details would you like?

winterz:
> > > This looks good, but test_builtin_compat_tzid() still fails for me.
> > 
> > 
> > can you provide details? the github CI is happy -- but we could be missing a combination of cmake options that could case a failure I suppose
> 
> @winterz What details would you like?

where it fails
what cmake options use used
platform(s)

ksuther:
I'm seeing the following when building on Mac:
```
not ok 1028 - Matches current TZID
# test failed: "(zone == tz)"
#          at: /Users/kent/Programming/libical/src/test/regression.c:4858
Returned builtin zone (NULL) doesn't match expected zone for TZID '/freeassociation.sourceforge.net/Europe/London'
not ok 1029 - Matches Tzfile compat TZID
# test failed: "(zone == tz)"
#          at: /Users/kent/Programming/libical/src/test/regression.c:4858
Returned builtin zone (NULL) doesn't match expected zone for TZID '/freeassociation.sourceforge.net/Tzfile/Europe/London'
not ok 1030 - Matches citadel.org TZID
# test failed: "(zone == tz)"
#          at: /Users/kent/Programming/libical/src/test/regression.c:4858
Returned builtin zone (NULL) doesn't match expected zone for TZID '/citadel.org/20190914_1/Europe/London'
```

Ran on a fresh build of this branch:
```
cmake .. -DENABLE_GTK_DOC=OFF
make regression
src/test/regression
```

ksmurchison:
Same errors on my Fedora linux box

cmake -DUSE_BUILTIN_TZDATA=true -DICAL_ALLOW_EMPTY_PROPERTIES=true

winterz:
I can't reproduce either failure.  could you have local modifications that might be causing the failures?

ksmurchison:
> I can't reproduce either failure. could you have local modifications that might be causing the failures?

No.  I'm using a clean pull of the tzid_prefix_fix branch.

PR #510, opened 2021-10-10
Add include path to the exported target definition

nicolasfella:
Otherwise the libical headers are not found by consumers if libical is installed in a non-standard location

### DIFF
--- a/src/libical/CMakeLists.txt
+++ b/src/libical/CMakeLists.txt
@@ -265,6 +265,8 @@ elseif(STATIC_ONLY)
   add_library(ical-static ALIAS ical)
 endif()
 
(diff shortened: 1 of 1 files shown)


src/libical/CMakeLists.txt
```diff
@@ -265,6 +265,8 @@ elseif(STATIC_ONLY)
   add_library(ical-static ALIAS ical)
 endif()
 
+target_include_directories(ical INTERFACE "$<INSTALL_INTERFACE:${INCLUDE_INSTALL_DIR}>")
```
nicolasfella, line 268: do we expect people to do 
`#include <libical/ical.h>`

or

`#include <ical.h>` ?

If it's the latter this needs to be

```suggestion
target_include_directories(ical INTERFACE "$<INSTALL_INTERFACE:${INCLUDE_INSTALL_DIR}>/libical")
```

PR #511, opened 2021-10-11
Make sure the libical target brings the includes directory

aleixpol:
We want that when we link against a target, its include directories come
with it. Make sure it's the case.


nicolasfella:
That looks a lot like https://github.com/libical/libical/pull/510

PR #519, opened 2021-11-08
icalcomponent.c: handle the case, when DTEND and DURATION are missing

dilyanpalauzov:
In icalcomponent_get_dtend():
• when both DTEND and DURATION are missing,
  set DTEND to DTSTART when the value-type of DTSTART is DATE-TIME,
  otherwise set DTEND = DTSTART + 1 day, when the value-type of DTSTART is DATE
• throw an error, if both DTEND and DURATION are present, as it is done in icalcomponent_get_duration()

In icalcomponent_get_duration():
• if DTEND and DURATION are missing and
  the value-type of DTSTART is DATE-TIME return zero duration;
  otherwise, if the value-type of DTSTART is DATE, return one day duration
• do not throw an error, if DTEND and DURATION are missing

In icalcomponent_get_span() account for the fact, that icalcomponent_get_dtend()
returns the correct DTEND, based on the value-type of DTSTART, when DTEND is missing.

Likewise for icalcomponent_foreach_recurrence().

Adjust test/regression-component.c to assume span.start == span.end,
when the event has only DTSTART and its value-type is DATE-TIME.

### DIFF
--- a/src/libical/icalcomponent.c
+++ b/src/libical/icalcomponent.c
@@ -726,24 +726,14 @@ or empty VCALENDAR component"); */
     }
     span.start = icaltime_as_timet_with_zone(start, icaltimezone_get_utc_timezone());
 
-    /* The end time could be specified as either a DTEND or a DURATION */
+    /* The end time could be specified as either a DTEND, a DURATION, or be missing */
     /* icalcomponent_get_dtend takes care of these cases. */
     end = icalcomponent_get_dtend(comp);
-    if (icaltime_is_null_time(end)) {
-        if (!icaltime_is_date(start)) {
-            /* If dtstart is a DATE-TIME and there is no DTEND nor DURATION
-               it takes no time */
-            span.start = 0;
-            return span;
-        } else {
-            end = start;
-        }
-    }
 
     span.end = icaltime_as_timet_with_zone(end, icaltimezone_get_utc_timezone());
     if (icaltime_is_date(start)) {
         /* Until the end of the day */
-        span.end += 60 * 60 * 24 - 1;
+        span.end -= 1;
     }
 
     return span;
@@ -891,16 +881,9 @@ void icalcomponent_foreach_recurrence(icalcomponent *comp,
     if (icaltime_is_null_time(dtstart))
         return;
 
-    /* The end time could be specified as either a DTEND or a DURATION */
+    /* The end time could be specified as either a DTEND, a DURATION or be missing */
     /* icalcomponent_get_dtend takes care of these cases. */
     dtend = icalcomponent_get_dtend(comp);
-    if (icaltime_is_null_time(dtend) && icaltime_is_date(dtstart)) {
-        /* No DTEND or DURATION and DTSTART is DATE - duration is 1 day */
-        struct icaldurationtype dur = icaldurationtype_null_duration();
-
-        dur.days = 1;
-        dtend = icaltime_add(dtstart, dur);
-    }
 
     /* Now set up the base span for this item, corresponding to the
        base DTSTART and DTEND */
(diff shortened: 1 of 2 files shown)


dilyanpalauzov:
In VAVAILABILITY when DTEND and DURATION are missing, the availability lasts forever, so different logic than for VEVENT applies.

dilyanpalauzov:
Actually, icalcomponent_get_duration() shall be set to
* DTSTART - DTEND for VEVENT, with fallback to 0/1 days when DURATION and DTEND are missing
* DTSTART - DUE for VTODO
* DTSTART - DTEND for VAVAILABILITY with fallback to infinite when DURATION and DTEND are missing

mcrha:
I just noticed that this change causes a regression in the evolution-data-server unit tests. The test-cal-reminders test fails to match a generated occurrence of the alarm from a component:
```
BEGIN:VTODO
UID:123
CREATED:20190829T000000Z
LAST-MODIFIED:20190829T000000Z
DTSTAMP:20190829T000000Z
SUMMARY:test
DTSTART:20190829T110000Z
DUE:20190830T100000Z
...
END:VTODO
```
because the returned "duration" is `0` in this case (the computed DTEND is equal to the DTSTART), while the previous behaviour was that the computed DTEND matched the DUE.

Is this new behaviour correct?

dilyanpalauzov:
The behaviour of icalcomponent_get_duration() is not correct.  It shall use DUE instead of DTEND for tasks.

mcrha:
Except the evolution-data-server uses icalcomponent_get_dtend().

dilyanpalauzov:
https://github.com/libical/libical/pull/534 corrects calcomponent_get_duration().

icalcomponent_get_dtend() is not defined on VTODO, it works on VEVENT and should work on VAVAILABILITY, as only the latter have DTEND property.

Before 94301bf8a152fefa icalcomponent_get_dtend() has returned icaltime_null_time(), when called on VTODO .

After 94301bf8a152fefa icalcomponent_get_dtend() returns the value of DTSTART, when called on VTODO or VEVENT, and the value type of DTSTART is DATE-TIME.  For VEVENT this behaviour is correct, for VTODO icalcomponent_get_due() shall be used instead.

dilyanpalauzov:
Shall `icalcomponent_get_dtend()` return `icaltime_null_time()`, when it is not called on VAVAILABILITY or VEVENT component?

mcrha:
Okay, I misplaced the DUE and DURATION, thinking of the developer documentation on the `icalcomponent_get_dtend`:
```
 *      For the icalcomponent routines only, dtend and duration are tied
 *      together. If you call the get routine for one and the other
 *      exists, the routine will calculate the return value. That is, if
 *      there is a DTEND and you call get_duration, the routine will
 *      return the difference between DTEND and DTSTART.
```

> Since 94301bf icalcomponent_get_dtend() returns the value of DTSTART, when called on VTODO or VEVENT, and the value type of DTSTART is DATE-TIME. For VEVENT this behaviour is correct,...

How can it be correct, when there is no DTEND (nor DURATION) property? It should return a "null" time, no?

dilyanpalauzov:
In VEVENT with DTSTART without DTEND, there is implicit value for DTEND.  The implicit value is =DTSTART, when the value type of DTSTART is date-time (zero seconds duration event).  The implicit value is =DTSTART +1day, if the value type of DTSTART is DATE.

For VTODO there is no such thing, as VTODO have no DTEND property.

mcrha:
Strictly speaking, the auto-compute of the DTEND without the DURATION property is not documented and the new behaviour breaks the code relying on that. It can also break the code, which somehow counts with the explicit things, and covers its own logic on the "implicit" DTEND in such case. I mean, similarly as one may expect to not get summary "calculated" when not present, it might not expect to get the DTEND calculated otherwise than documented.

Nonetheless, I'm going to fix the evolution-data-server code, at least the parts I noticed, to work the same with and without this change.

mcrha:
> Shall `icalcomponent_get_dtend()` return `icaltime_null_time()`, when it is not called on VAVAILABILITY or VEVENT component?

I think this makes sense. The current behaviour pretends the DTEND exists, even you repeatedly confirmed the DTEND makes sense only for the VEVENT and VFREEBUSY, which is declared in the https://datatracker.ietf.org/doc/html/rfc5545#section-3.8.2.2 .

mcrha:
> The auto-computing of DTEND for VEVENTS without DURATION and without DTEND, is documented in RFC5545:

I understand it as how the application should behave, which doesn't necessarily mean the underlying library should force it. Sometimes you want to check whether certain property exists simply by reading it, instead of using proper function for it, because the library in question behaves as expected by the code.

I think this matter is solved now, no need to chat about it. I do agree with you that the evolution-data-server/evolution code did not read the properties only when it could, which your change uncovered.

PR #520, opened 2021-11-08
restrictions.csv: DECLINECOUNTER,VEVENT: DURATION and DTEND are mutually exclusive

dilyanpalauzov:
My reading of [RFC 5546 - 3.2.8](https://datatracker.ietf.org/doc/html/rfc5546#section-3.2.8) is that for METHOD:DECLINECOUNTER either DTEND or DURATION must be present.  If both are absent, the usual logic of calculating the end
applies, based on the value-type of DTSTART.


winterz:
@dilyanpalauzov thanks for all your recent patches.  I don't think I'll have a chance to review until the weekend, fyi.  Happy for @ksuther or @ksmurchison to take a look in the meantime.

I am wondering if these patches belong in the 3.0 (stable) branch as well.

ksuther:
I'm not an expert on the restrictions.csv format, but conceptually that sounds right.

ksmurchison:
I'm returning from vacation right now. I'll take a look in the morning

PR #526, opened 2021-11-18
doc/UsingLibical.md – typos

dilyanpalauzov:


winterz:
why do we still have both the .md and the .txt versions anyway
I'm going to trash the .txt version

PR #534, opened 2022-01-04
icalcomponent_get_duration() — calculate the duration differently for VEVENT and VTODO

dilyanpalauzov:
For VEVENT and VAVAILABILITY calculate the distance between DTSTART and DTEND.

For VTODO calculate the distance between DTSTART and DUE.

icalcomponent_get_dtend() — return null-time, unless called on VEVENT, VAVAILABILITY or VFREEBUSY.

### DIFF
--- a/src/libical/icalcomponent.c
+++ b/src/libical/icalcomponent.c
@@ -1406,9 +1406,22 @@ struct icaltimetype icalcomponent_get_dtstart(icalcomponent *comp)
 struct icaltimetype icalcomponent_get_dtend(icalcomponent *comp)
 {
     icalcomponent *inner = icalcomponent_get_inner(comp);
-    icalproperty *end_prop = icalcomponent_get_first_property(inner, ICAL_DTEND_PROPERTY);
-    icalproperty *dur_prop = icalcomponent_get_first_property(inner, ICAL_DURATION_PROPERTY);
+    const icalcomponent_kind kind = icalcomponent_isa(inner);
+    icalproperty *end_prop, *dur_prop;
     struct icaltimetype ret;
+
+    switch(kind) {
+    case ICAL_VAVAILABILITY_COMPONENT:
+    case ICAL_VEVENT_COMPONENT:
+    case ICAL_VFREEBUSY_COMPONENT:
+    case ICAL_XAVAILABLE_COMPONENT:
+        break;
+    default:
+        return icaltime_null_time();
+    }
+
+    end_prop = icalcomponent_get_first_property(inner, ICAL_DTEND_PROPERTY);
+    dur_prop = icalcomponent_get_first_property(inner, ICAL_DURATION_PROPERTY);
 
     if (end_prop != 0 && dur_prop == 0) {
         ret = icalproperty_get_datetime_with_component(end_prop, comp);
@@ -1426,13 +1439,17 @@ struct icaltimetype icalcomponent_get_dtend(icalcomponent *comp)
 
         ret = icaltime_add(start, duration);
     } else if (end_prop == 0 && dur_prop == 0) {
-        struct icaltimetype start = icalcomponent_get_dtstart(inner);
-        if (icaltime_is_date(start)) {
-            struct icaldurationtype duration = icaldurationtype_null_duration();
-            duration.days = 1;
-            ret = icaltime_add(start, duration);
+        if (kind == ICAL_VEVENT_COMPONENT) {
+            struct icaltimetype start = icalcomponent_get_dtstart(inner);
+            if (icaltime_is_date(start)) {
+                struct icaldurationtype duration = icaldurationtype_null_duration();
+                duration.days = 1;
+                ret = icaltime_add(start, duration);
+            } else {
+                ret = start;
+            }
         } else {
-            ret = start;
+            ret = icaltime_null_time();
         }
     } else {
         /* Error, both duration and dtend have been specified */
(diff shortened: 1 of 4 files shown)


winterz:
@dilyanpalauzov could you also add a blurb to the top-level ReleaseNotes.txt about this change?  Maybe add a test to regression.c too

@mcrha to double-check: you are ok with this patch now?

winterz:
nit: 2 lines in your patch are longer than 100 chars.  would be appreciated if you'd split those over more than 1 line.

mcrha:
> @mcrha to double-check: you are ok with this patch now?

Yes, I am. It makes the evolution-data-server tests pass without additional changes (but I'll keep the changes there, because they make sense).

src/libical/icalcomponent.c
```diff
@@ -1406,6 +1406,12 @@ struct icaltimetype icalcomponent_get_dtstart(icalcomponent *comp)
 struct icaltimetype icalcomponent_get_dtend(icalcomponent *comp)
 {
     icalcomponent *inner = icalcomponent_get_inner(comp);
+    const icalcomponent_kind kind = icalcomponent_isa(inner);
+    if (kind != ICAL_VEVENT_COMPONENT && kind != ICAL_VAVAILABILITY_COMPONENT
```
ksmurchison, line 1410: I think a switch() statement would be cleaner (and more extendible) than an if.  I need to tweak the VPOLL support to match the current IETF draft, and will have to add it in here eventually.

src/libical/icalcomponent.c
```diff
@@ -1505,12 +1524,13 @@ struct icaldurationtype icalcomponent_get_duration(icalcomponent *comp)
          * The standard actually allows different time zones.
          */
         struct icaltimetype start = icalcomponent_get_dtstart(inner);
-        struct icaltimetype end = icalcomponent_get_dtend(inner);
+        struct icaltimetype end = kind == ICAL_VTODO_COMPONENT
```
ksmurchison, line 1527: Can't we grab the end time in the new switch statement above?

ksmurchison (changes requested):
The logic looks good to me, but I have a couple of comments inline.

dilyanpalauzov:
I have added text to ReleaseNotes.txt, added tests to regression.c, switched to `switch()`, extended the documentation of `icalcomponent_get_duration()` and `icaltimetype icalcomponent_get_dtend()` about the implicit calculations for VEVENT.

dilyanpalauzov:
I added handling for the AVAILABILITY component, apart from VTODO, VEVENT, VFREEBUSY and VAVAILABILITY components.

dilyanpalauzov:
I wrote `AVAILABILITY`, but I meant `AVAILABLE`.

winterz:
I think we need this in the 3.0 branch as well.  The patch doesn't apply cleanly to 3.0, but I can cherry-pick and fix it, hopefully.
Any other final comments @ksmurchison or @mcrha ?

ksmurchison (approved):

ksmurchison (approved):
LGTM

src/libical/icalcomponent.c
```diff
@@ -1406,6 +1406,18 @@ struct icaltimetype icalcomponent_get_dtstart(icalcomponent *comp)
...
+    case ICAL_VAVAILABILITY_COMPONENT:
+    case ICAL_VEVENT_COMPONENT:
+    case ICAL_VFREEBUSY_COMPONENT:
+    case ICAL_XAVAILABLE_COMPONENT:
+        break;
+    default:
+        return icaltime_null_time();
+    }
+
     icalproperty *end_prop = icalcomponent_get_first_property(inner, ICAL_DTEND_PROPERTY);
```
winterz, line 1421: declare `icalproperty *end_prop, *dur_prop` and `struct icaltimetype ret` above the switch else some older C compilers will barf.

src/libical/icalcomponent.c
```diff
@@ -1488,12 +1504,32 @@ void icalcomponent_set_duration(icalcomponent *comp, struct icaldurationtype v)
...
+        if (end_prop) {
+            end = icalcomponent_get_due(inner);
+        }
+        break;
+    default:
+        /* The libical API is used incorrectly */
+        return icaldurationtype_null_duration();
+    }
 
     icalproperty *dur_prop = icalcomponent_get_first_property(inner, ICAL_DURATION_PROPERTY);
```
winterz, line 1531: same here. `icalproperty *dur_prop` and `struct icaldurationtype ret` must be declared above the switch

winterz (commented):
fixes needed for old C compilers

dilyanpalauzov:
I removed the handling of AVAILABLE, and left VAVAILABILITY.   The `icalcomponent_get_duration()` and `icalcomponentget_dtend()` functions get as parameter a VCALENDAR component and act on its first child component. AVAILABLE is not direct child of VCALENDAR.

mcrha:
> functions get as parameter a VCALENDAR component and act on its first child component. AVAILABLE is not direct child of VCALENDAR.

Passing a VCALENDAR component feels like a misuse of the function. You should not expect anything about the caller's intention when he/she does so. The first child component can be a VTIMEZONE or any other. I mean, I would not use the first child, when the passed in component is not one of the expected component types, just return a null-time.

dilyanpalauzov:
I was wrong: `icalcomponent_get_duration()` and `icalcomponet_get_dtent()` call `icalcomponent_get_inner()`.  For VCALENDAR `icalcomponent_get_inner()` returns the first child, otherwise it returns the passed parameter.

I have restored the handling of the AVAILABLE component.

dilyanpalauzov:
Removing `icalcomponent *inner = icalcomponent_get_inner(comp);` from `icalcomponent_get_dtend()` and others shall be handled in a different ticket.

mcrha:
> Removing `icalcomponent *inner = icalcomponent_get_inner(comp);` from `icalcomponent_get_dtend()` and others shall be handled in a different ticket.

I agree. I understood from your previous comment that you added that there. If it was there already, then it can be some library users depend on the current behavior. In that case it's safer to be left there, even I still believe it's a bad idea.

winterz:
I'll work on cherry-picking this back to the 3.0 branch.
unless there are objections?

mcrha:
No objection from my side. The evolution-data-server unit tests pass with it with no problem (the apply of the patch reports a lot of offset-s, but otherwise applies the code changes cleanly).

PR #539, opened 2022-01-20
Fix icalcomponent_set_due not removing TZID when necessary

ksuther:
Stumbled upon a mismatch in behavior between `icalcomponent_set_due` and `icalcomponent_set_dtstart`. `icalcomponent_set_dtstart` strips TZID correctly, but `icalcomponent_set_due` doesn't.

### DIFF
--- a/src/libical/icalcomponent.c
+++ b/src/libical/icalcomponent.c
@@ -2471,6 +2471,7 @@ void icalcomponent_set_due(icalcomponent *comp, struct icaltimetype v)
         icalcomponent_add_property(inner, due_prop);
     } else if (due_prop != 0) {
         icalproperty_set_due(due_prop, v);
+        icalproperty_remove_parameter_by_kind(due_prop, ICAL_TZID_PARAMETER);
     } else if (dur_prop != 0) {
         struct icaltimetype start = icalcomponent_get_dtstart(inner);
 
--- a/src/test/regression.c
+++ b/src/test/regression.c
@@ -5169,6 +5169,27 @@ test_icalvalue_resets_timezone_on_set(void)
     icalerror_clear_errno();
 }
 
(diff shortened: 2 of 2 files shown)


winterz:
should go into the 3.0 branch too, right?

winterz (approved):
I'll cherry-pick this into the 3.0 branch after you merge.

src/libical/icalcomponent.c
```diff
@@ -2461,8 +2461,12 @@ void icalcomponent_set_due(icalcomponent *comp, struct icaltimetype v)
         icalproperty_set_duration(dur_prop, dur);
     }
 
-    if (due_prop && (tzid = icaltime_get_tzid(v)) != NULL && !icaltime_is_utc(v)) {
-        icalproperty_set_parameter(due_prop, icalparameter_new_tzid(tzid));
+    if (due_prop) {
+        if ((tzid = icaltime_get_tzid(v)) != NULL && !icaltime_is_utc(v)) {
+            icalproperty_set_parameter(due_prop, icalparameter_new_tzid(tzid));
+        } else {
+            icalproperty_remove_parameter_by_kind(due_prop, ICAL_TZID_PARAMETER);
```
ksmurchison, line 2468: Not that its critical, but couldn't the remove_param line simply be added after line #2453?  This would more closely match what set_dtstart looks like

ksuther: You're right. I initially didn't do that because it seemed like calling `icalproperty_remove_parameter_by_kind` every time was unnecessary, but it's better with your suggestion.

ksuther:
Rebased to get rid of conflicts. Makes sense for this to go into 3.0 as well.

winterz:
master and 3.0 with 90320c9b4383d6a766f493fd5b22779fd927d21a

PR #540, opened 2022-01-21
Consistent use of icalmemory_new_buffer et al.

minichma:
In the libical core code the use of functions from `icalmemory.h` is somewhat inconsistent. In some places memory is managed using the functinos from icalmemory while in others the functions from stdlib (malloc/free, etc.) are used and in some places the usage is mixed. This PR replaces the use of memory management functions from stdlib with equivalents from `icalmemory.h` to improve overall consistency, at least in the code under `src/libical`. I.e. the following functions are replaced

| old     | new |
| --------- | --- | 
| calloc()  | icalmemory_new_buffer() |
| malloc()  | icalmemory_new_buffer() |
| realloc() | icalmemory_resize_buffer() |
| free()    | icalmemory_free_buffer() |
| strdup()  | icalmemory_strdup() |

Besides making the code more consistent, this also has the following benefits:
* Projects consuming libical as source code can more easily replace the memory management functions if necessary. This especially makes sense for code running on embedded systems, where malloc et al. might not be available.
* `icalmemory.h` could easily be extended such that memory management functions could also be replaced at runtime. I would actually contribute such code in a subsequent PR.
* The memory management functions could easily be replaced in the automated tests in order to detect memory leaks and inconsistent use of memory functions. I would also contribute such in a subsequent PR.

As the functions in icalmemory still make use of the same functions from stdlib internally, this PR doesn't introduce any functional change. I.e. clients which today are freeing libical-allocated memory using `free()` can safely continue to do so.

### DIFF
--- a/src/libical/icalrecur.c
+++ b/src/libical/icalrecur.c
@@ -593,7 +593,7 @@ static int icalrecur_add_bydayrules(struct icalrecur_parser *parser,
         icalrecurrencetype_weekday wd;
 
         if (i == ICAL_BY_DAY_SIZE) {
-            free(vals_copy);
+            icalmemory_free_buffer(vals_copy);
             return -1;
         }
 
@@ -624,7 +624,7 @@ static int icalrecur_add_bydayrules(struct icalrecur_parser *parser,
 
         /* Sanity check value */
         if (wd == ICAL_NO_WEEKDAY || weekno >= ICAL_BY_WEEKNO_SIZE) {
-            free(vals_copy);
+            icalmemory_free_buffer(vals_copy);
             return -1;
         }
 
@@ -632,7 +632,7 @@ static int icalrecur_add_bydayrules(struct icalrecur_parser *parser,
         array[i] = ICAL_RECURRENCE_ARRAY_MAX;
     }
 
-    free(vals_copy);
+    icalmemory_free_buffer(vals_copy);
 
     sort_bydayrules(parser);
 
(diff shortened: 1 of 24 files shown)


src/libical/icalrecur.c
```diff
@@ -2026,7 +2026,7 @@ icalrecur_iterator *icalrecur_iterator_new(struct icalrecurrencetype rule,
                 impl->orig_data[byrule] = 0;
             } else {
                 icalerror_set_errno(ICAL_MALFORMEDDATA_ERROR);
-                free(impl);
+            icalmemory_free_buffer(impl);
```
winterz, line 2029: indentation

src/libicalss/icalclassify.c
```diff
@@ -292,8 +292,8 @@ int icalssutil_is_rescheduled(icalcomponent *a, icalcomponent *b)
         temp1 = icalproperty_as_ical_string_r(p1);
         temp2 = icalproperty_as_ical_string_r(p2);
         cmp = strcmp(temp1, temp2);
-        free(temp1);
-        free(temp2);
+        icalmemory_free_buffer(temp1);
+		icalmemory_free_buffer(temp2);
```
winterz, line 296: remove that leading tab while we're at it

src/libicalss/icalfileset.c
```diff
@@ -397,11 +398,11 @@ icalerrorenum icalfileset_commit(icalset *set)
         if (sz != (IO_SSIZE_T) strlen(str)) {
             perror("write");
             icalerror_set_errno(ICAL_FILE_ERROR);
-            free(str);
+            icalmemory_free_buffer(str);
             return ICAL_FILE_ERROR;
         }
 
-        free(str);
+		icalmemory_free_buffer(str);
```
winterz, line 405: another leading tab

src/libical/icaltimezone.c
```diff
@@ -28,6 +28,7 @@
 #include "icalparser.h"
 #include "icalmemory.h"
 #include "icaltz-util.h"
+#include "icalmemory.h"
```
winterz, line 31: dupe.  we already have this 2 lines up

src/libical/icalrecur.c
minichma: sorry for that - classical rebasing issues

PR #541, opened 2022-01-21
Centrally #define `fprintf` and `assert` to allow easy replacement.

minichma:
This PR centralizes the use of `fprintf(err, ...)` and `assert()` within `src/libical` to allow easy replacement where necessary. E.g. when compiling for resource-constrained devices, linking `fprintf()` might not be desirable due to the imposed program size. By using centrally defined macros instead, the functions can easily be replaced by no-ops. We could also easily add flags for this to the CMAKE config in a future PR if deemed required.

### DIFF
--- a/config.h.cmake
+++ b/config.h.cmake
@@ -548,3 +548,6 @@ typedef ssize_t IO_SSIZE_T;
 #define _unused(x) (void)x
 #endif
 #endif
+
+#define icalassert(...) assert(__VA_ARGS__)
+#define icalerrprintf(...) fprintf(stderr, __VA_ARGS__)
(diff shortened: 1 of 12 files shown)


config.h.cmake
```diff
@@ -548,3 +548,6 @@ typedef ssize_t IO_SSIZE_T;
 #define _unused(x) (void)x
 #endif
 #endif
+
+#define icalassert(...) assert(__VA_ARGS__)
```
minichma, line 552: @winterz Is this the right place for defining those macros?

winterz: yep

winterz: we need a rebase. there are conflicts..

minichma: here it comes

winterz:
"Variadic macros became a standard part of the C language with C99."  
that may rule out some of the more ancient C compilers we've been supporting in the 3.0 branch.
not that I mind.  I think we can start requiring C99 by now :)

PR #542, opened 2022-01-22
Make icalmemory & icalarray more robust against out of memory

minichma:
This PR makes some functions from icalmemory, icalarray and some other files somewhat more robust against out of memory errors. This is especially, but not only relevant for resource-constrained devices like embedded devices. The PR covers just a few functions I came across but there's certainly more to find. It would probably be a good idea to do a more comprehensive analysis as errors in memory management could easily cause security issues.


minichma:
@winterz what do you think?

winterz:
looks good.  thank you for these improvements.

PR #546, opened 2022-02-18
buildsystem - entire project can be built with memory consistency

winterz:
Introduce a new CMake option LIBICAL_DEVMODE_MEMORY_CONSISTENCY
that allows all the project libraries to be built with the
memory consistency functions.


winterz:
@minichma take a look at this patch please.

try it with
```cmake -DLIBICAL_DEVMODE_MEMORY_CONSISTENCY .. && make && make test
```
then you'll notice the `process` and `icalrecurtest-r` dump core.
not sure I'm doing things exactly right yet.

let me know what you think.

minichma:
> then you'll notice the process and icalrecurtest-r dump core.

This seems to be caused by an actual inconsistency at 

https://github.com/libical/libical/blob/0f33ebaff96143b2ab9001f3841f1c51c8cc2814/src/libicalss/icalmessage.c#L81

The regular `free()` function is used with memory allocated via icalmemory. The new flag seems to be working. :-)

Didn't have time to debug in more detail (took me some time to get the debugger up and running on Linux, the problem doesn't show up on Windows). Will hopefully find the time to have a closer look later today or tomorrow.

minichma:
Btw. it would be great if we managed to activate the new check during build checks that are performed on PRs. Is there an easy way of enabling this?

minichma:
I added fixes for the failing tests. The issue was, that  the use of icalmemory is not implemented throughout whole libical. In my previous memory consistency PR I mainly looked at the files in src/libical so far. Not using it everywhere is basically fine, it just should be used in a consistent manner. It's getting tricky, however, where libical frees memory that was allocated by the user or vice versa. I'm not sure, whether this is something allowed to happen. At least when dealing with rscale this seems to be the case. I.e. the data pointed at by `icalrecurrencetype.rscale` is usually allocated within `icalrecur.c` but usually freed outside. This is what caused the icalrecurtest-r test to fail.

However, I think, this is something we can look at separately. At least we have the tools now for doing so :-)

Btw. the issue with icalrecurrencetype, which is a value type, pointing at allocated memory, is something I stumbled upon in other cases too. Would it maybe make sense to allocate icalrecurrencetype dynamically (via some icalrecur_new() function) and free everything via a single icalrecur_free() function? This would also allow us to make the whole struct more dynamic. I'll create a ticket for discussing this.

src/libical/icalmemory.c
```diff
@@ -281,19 +284,25 @@ char *icalmemory_strdup(const char *s)
     return res;
 }
 
-#if defined(ICALMEMORY_DEFAULT_MALLOC) && !defined(S_SPLINT_S)
+#if defined(MEMORY_CONSISTENCY)
+static icalmemory_malloc_f global_icalmem_malloc = &test_malloc;
```
minichma, line 288: Would it make sense to make the distinction between test versions and regular versions in config.h.cmake? Then we wouldn't have to deal with MEMORY_CONSISTENCY details here. But then we'd have to think about how to `#include "test-malloc.h"`. Not sure which way'd be better.

minichma (changes requested):
Looks good to me. Gives us a nice little tool at hand. If possible, we should activate the flag for verification builds (branch/PR). Only thing is the commented code left over. And we could think about moving the distinction between malloc and test_malloc out to config.h.cmake. But not sure, whether this makes sense at all.

src/libical/icalmemory.c
winterz: yes, I started moving test-malloc directly into libical  but the patch because a mess.

src/test/test-malloc.c
```diff
@@ -146,8 +145,8 @@ void test_free(void *p) {
         //   icalmemory, e.g. via malloc().
         // * The header in front of the memory block being freed has been corrupted.
 
-        ok("freed memory was allocated via icalmemory and has not been corrupted",
-           hdr->magic_no == TESTMALLOC_MAGIC_NO);
+        //ok("freed memory was allocated via icalmemory and has not been corrupted",
```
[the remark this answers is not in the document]
winterz: I used a conditional compile instead.  seems useful to keep the ok() message

winterz:
> Btw. it would be great if we managed to activate the new check during build checks that are performed on PRs. Is there an easy way of enabling this?

Yes should be easy enough

winterz:
I think I'll commit the latest version of this patch. then tweak as needed.

PR #549, opened 2022-02-23
Substitute time_t with centrally defined icaltime_t to make dealing with 32-bit restrictions easier

minichma:
`time_t` and related functions are system-defined and might vary among different architectures. Most notably, the range of time_t ends in the year 2038 on 32-bit architectures. This PR substitutes time_t and related functions by the newly introduced `icaltime_t`, which is centrally defined in the new `config_public.h.cmake` and `config.h.cmake`. By default `icaltime_t` is an alias for `time_t`, so the change is fully backwards-compatible. However, by centrally defining it, it can easily be replaced with custom types and functions where required.

The new `config_public.h.cmake` file was required, because some types that are included in `ical.h` refer to the new `icaltime_t` but `config.h` is obviosly considered private and therefore not part of `ical.h`. In contrast, the content of `config_public.h` is also included in `ical.h` and can therefore be used from within the `ical*.h` files.

Future work:
* `gmtime_r` and `mktime` could be replaced with libical-own functionality quite easily (i.e. by using caldat_int, etc.). This would make libical even more independent from system-defined functions.
* Compilers like GCC and MSVC provide 64-bit versions of time_t even on 32-bit architectures. We could add a cmake option for using those instead of the standard versions.


thesamesam:
Thanks for caring about this kind of use case. It's important to us in distributions.

I'm wondering whether it's worth catering to systems which _don't_ end up migrating to a 64-bit `time_t`. [musl has already made the migration with >= 1.2.0](https://musl.libc.org/time64.html) and glibc started the process with 2.34. We started writing up some of our [plans](https://wiki.gentoo.org/wiki/Project:Toolchain/time64_migration) in Gentoo for 32-bit architectures to get everyone rebuilt with the needed new definitions.

I know [OpenBSD](https://www.openbsd.org/papers/eurobsdcon_2013_time_t/) already cut the cord too, as did NetBSD. Not sure about FreeBSD.

What I'm saying is: I'm not sure what useful platforms are actually stuck on 32-bit time_t, as most (all?) 32-bit common targets should be able to rebuild with new enough musl, new enough glibc, or new enough BSD, and get a new shiny time_t, and if there are any not in this category, does libical want/need to support them?

Not opposing at all, just wanted to point out that this might solve itself, and could avoid some code complexity in the process. But I'm very happy to see this and if it's not relevant, do ignore me! Thanks for caring about portability. It might be I'm missing some platforms you support too :)

minichma:
@thesamesam thanks for the insights! It's good to see, projects are moving towards 64-bit time_t. GCC (and probably others too) still defaults to 32-bit time_t on 32-bit systems though. In the kind of downstream projects, where the detailed compiler settings are not under  control of the authors, this might cause issues. E.g. here at Tapkey we deliver code to partners that integrate it into their embedded hardware (most often 32-bit ARM Cortex-M). It's up to them which compilers and settings they use. However, it's up to us to ensure, our code won't stop working in 2038. So allowing decoupling icaltime_t from the downstream compilers and settings significantly reduces overall complexity in such scenarios. Considering the rather minor additional complexity that comes from the additional abstraction, I feel it's probably worth it.

winterz:
> @winterz I didn't handle references to time_t in the wrapers and API definitions for Python, Net, Glib, because I'm not really into those. Should not be necessary for now, because all changes are backwards-compatible. What do you think?
> 

I think we are ok on those wrappers for now.  We haven't maintained the Python or Perl wrappers at all.

> `SIZEOF_ICALTIME_T` is simply defined as `#define SIZEOF_ICALTIME_T SIZEOF_TIME_T`. How would this be defined automatically by cmake?

We already have the CMake option `USE_32BIT_TIME_T`.    We could add `USE_64BIT_TIME_T` ?

winterz:
we are (potentially) changing the signatures of some public functions that use time_t.
I think we need to deprecate and add new functions.

deprecate these:
- icalrecur_expand_recurrence()
- icaltime_from_timet_with_zone()
- icaltime_as_timet()
- icaltime_as_timet_with_zone()
add these:
- icalrecur_expand_recurrence_to_icaltimes()
- icaltime_from_icaltimet_with_zone()
- icaltime_as_icaltimet()
- icaltime_as_icaltimet_with_zone()

Also mention these API changes in the ReleaseNotes.txt

winterz:
do we need config_public.h.cmake? 
what if we change src/libical/icaltime.h into src/libical/icaltime.h.cmake and generate instead?
feels a bit cleaner, then all the icaltime stuff will be in one header.

minichma:
> we are (potentially) changing the signatures of some public functions that use time_t. I think we need to deprecate and add new functions.
> 
> deprecate these:
> 
> * icalrecur_expand_recurrence()
> * icaltime_from_timet_with_zone()
> * icaltime_as_timet()
> * icaltime_as_timet_with_zone()
> 
> add these:
> 
> * icalrecur_expand_recurrence_to_icaltimes()
> * icaltime_from_icaltimet_with_zone()
> * icaltime_as_icaltimet()
> * icaltime_as_icaltimet_with_zone()
> 
> Also mention these API changes in the ReleaseNotes.txt

As by default the new `icaltime_t` is just an alias for `time_t`, the signatures stay unchanged, as long as `icaltime_t` isn't intentionally defined differently. Also semantically nothing changes, so I thought, the function names could stay the same. Or do you think, we should update the function names, just because the type name is part of them?

minichma:
> do we need config_public.h.cmake? what if we change src/libical/icaltime.h into src/libical/icaltime.h.cmake and generate instead? feels a bit cleaner, then all the icaltime stuff will be in one header.

We don't necessarily need it. At this point nothing is even generated within config_public.h. I just thought it would be a good idea to design it in a way symmetric to the existing config.h.cmake, with the only difference, that it's included in ical.h. Defining `icaltime_t` inside `icaltime.h` would certainly be an option. I guess, this would require defining the underlying type in the top `CMakeLists.txt`, so we have it available during code generation, right? Would it then maybe be a good idea to split generated types out into a separate header file? Maybe `icaltypes.h.cmake`? Then we wouldn't have to mix generated types with regular code.

I'll make a suggestions.

minichma:
> We already have the CMake option USE_32BIT_TIME_T. We could add USE_64BIT_TIME_T ?

Yes, I think that's a good idea. As far as I understand the former, it's about the definition of `time_t` itself, whereas the latter would be about redirecting `icaltime_t` to a 64-bit type. So I suggest naming it `USE_64BIT_ICALTIME_T `.

If set, we would redirect `icaltime_t` and related functions to `__time64_t` and related on some well-known platforms (GCC, MSVC) while issuing a cmake-time error on other platforms?

I'll make a suggestion.

minichma:
Introduced new cmake option `USE_64BIT_ICALTIME_T`. Only implemented the redirects for MSVC for now as I don't have the time for properly implementing it for more compilers/platforms. But MSVC by default uses a 64-bit time_t anyways, except if the `_USE_32BIT_TIME_T` define is set, which we support via the `USE_32BIT_TIME_T` cmake option. But this option doesn't seem to be fully functional, as `SIZEOF_SIZE_T` is set to 8 even if the option is enabled (see https://github.com/libical/libical/issues/557). So for now the new option isn't of much use, which is why I marked it as 'experimental'. However, I'm not sure, whether we should keep the new option in this PR at all or rather implement it in a separate PR in a more complete manner. I believe, the redirects themselves (icaltime_t instead of time_t, ...) on the other hand *do* make sense as they are already, so I think, this PR makes sense even without fully implementing `USE_64BIT_ICALTIME_T`. What do you think?

I also removed config_public.h.cmake in favor of `icaltime.h.cmake` but I'm not sure, whether it's the cleanest approach to mix the source code of icaltime.h with stuff generated by cmake. Should we maybe introduce some `icaldefines.h.cmake` or the like and leave `icaltime.h` as it is?

winterz:
@minichma FYI
in commit 4f22bbb2d1d1f3a7a218e2d8d363f94a071151c9 I changed the CMake option from `USE_64BIT_ICALTIME_T ` to `LIBICAL_ENABLE_64BIT_ICALTIME_T`

PR #550, opened 2022-02-23
co-locate the MEMORY_CONSISTENCY settings in icalmemory.c

winterz:
Put all the possible settings for the MEMORY_CONSISTENCY in 1 place, icalmemory.c

### DIFF
--- a/config.h.cmake
+++ b/config.h.cmake
@@ -551,7 +551,3 @@ typedef ssize_t IO_SSIZE_T;
 
 #define icalassert(...) assert(__VA_ARGS__)
 #define icalerrprintf(...) fprintf(stderr, __VA_ARGS__)
-
-#define ICALMEMORY_DEFAULT_MALLOC malloc
-#define ICALMEMORY_DEFAULT_REALLOC realloc
-#define ICALMEMORY_DEFAULT_FREE free
--- a/scripts/buildtests.sh
+++ b/scripts/buildtests.sh
@@ -332,9 +332,6 @@ CPPCHECK() {
            -D PTHREAD_RECURSIVE_MUTEX_INITIALIZER_NP=0 \
            -D _unused="(void)" \
            -D _deprecated="(void)" \
-           -D ICALMEMORY_DEFAULT_FREE="free" \
-           -D ICALMEMORY_DEFAULT_MALLOC="malloc" \
-           -D ICALMEMORY_DEFAULT_REALLOC="realloc" \
            -D F_OK=0 \
            -D R_OK=0 \
            -U YYSTYPE \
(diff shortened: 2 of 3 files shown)


winterz:
@minichma how do you feel about this?  since there were no other consumers for the ICALMEMORY_DEFAULT_* macros I don't see a reason to have them in config.h.cmake

minichma:
Having the settings in config.h would allow compile-time replacement without the need for changing the 'actual' code base. Here at Tapkey we use libical on embedded controllers (e.g. on Nordic's nRF52), where program size is a major concern. Replacing malloc at runtime is therefore not sufficient, as we also need to avoid linking it. By having malloc configured as default inside icalmemory, replacing would require us to fork the code base and patch icalmemory.c. By configuring the defauls in config.h, changing the code base would be avoided. However, in this very use case, centrally defining the defaults in config.h would be sufficient, whereas having top-level cmake settings would not be required.

That said, the described use case might not be very common and forking might be required anyways is such cases. So the benefit of improved configurability might not be worth the additional code and complexity.

As an intermediate solution we could centrally define the defaults in config.h.cmake but without having top-level cmake settings.

winterz:
I certainly don't want to make any changes that require a fork on your part.
Ideally you'd only need to change your CMake and/or CCFLAGS/LDFLAGS.

Is the current code good enough to avoid forking, or is more work needed?

I assume you can compile with whatever -DICALMEMORY_DEFAULT_* values you want the way things are right now.
we could also add CMake options for LIBICAL_USE_MEMORY_SYSTEM_BOO and change config.h.cmake, for example:
```
#if(LIBICAL_USE_MEMORY_SYSTEM_BOO)
#define ICALMEMORY_DEFAULT_MALLOC boo_malloc
#define ICALMEMORY_DEFAULT_REALLOC boo_realloc
#define ICALMEMORY_DEFAULT_FREE boo_free
#elif(LIBICAL_USE_MEMORY_SYSTEM_GOO)
#define ICALMEMORY_DEFAULT_MALLOC goo_malloc
#define ICALMEMORY_DEFAULT_REALLOC goo_realloc
#define ICALMEMORY_DEFAULT_FREE goo_free
#else
#define ICALMEMORY_DEFAULT_MALLOC malloc
#define ICALMEMORY_DEFAULT_REALLOC realloc
#define ICALMEMORY_DEFAULT_FREE free
#endif
```
or some-such

minichma:
Yes, the current (`master`) version certainly is good enough, because it allows us defining `malloc` as empty, therefore avoiding compiling and linking it into libical. From our side it would also be ok to have a single cmake option like 'ICAL_CUSTOM_MALLOC` that, if true, doesn't set a default at all and instead requires the user to set custom functions.

winterz:
abandoning this.
will look at adding a CMake option for ICAL_CUSTOM_MALLOC

PR #567, opened 2022-04-24
Use README.md as doxygen mainpage

### LINKED ISSUE #565: Embed README.md in the Mainpage.dox
In the Mainpage.dox can we embed (not a link, but the contents) the README.md?

rather than having to duplicate the same info in 2 places

CC: @mcclurgm

mcclurgm:
Fixes #565 

This is a WIP. I can't get doxygen to recognize any files in the base directory, only in the `doc` directory. I'm very unfamiliar with cmake, so I would appreciate any help getting this to work.

This currently builds the doxygen homepage based on the new `doc/README.md` file. I modified that file from the original `README` to work better as a doxygen homepage. I did my best to make it work as both a homepage for the GitHub repo and the documentation website, but it could probably be improved.

I can't get it to work without the `{#mainpage}` tag, which makes the doxyfile change sugested in #565 redundant. If that changes when fixing the build, then I would prefer to remove this tag for GitHub compatibility.

Once I get the build figured out I intend to replace the current `README.MD` with the contents of the new file for merge.


winterz:
@mcclurgm is this still a draft?  
what's the status

winterz:
superceeded by #568

PR #570, opened 2022-04-27
parameters.csv: Fix enum value clashing

mcrha:
Two problems:
* EMAIL being defined twice
* REASON and REQUIRED use the same value

The effect of this patch is that only the ICAL_REASON_PARAMETER
value changes from 43 to 47, the ICAL_EMAIL_PARAMETER value
stays unchanged.

### DIFF
--- a/design-data/parameters.csv
+++ b/design-data/parameters.csv
@@ -51,15 +51,14 @@
 "SIZE","41","const char*",
 "FILENAME","42","const char*",
 "#Task Extension Parameters","draft-apthorp-ical-tasks"
-"REASON","43","const char*",
+"REASON","47","const char*",
 "MODIFIED","44","const char*",
 "SUBSTATE","45","icalparameter_substate",X=21900;OK;ERROR;SUSPENDED;NONE=21999",
 "#Parameters from New Properties for iCalendar","RFC 7986 Section 6"
 "DISPLAY","46","icalparameter_display",X=22000;BADGE;GRAPHIC;FULLSIZE;THUMBNAIL;NONE=22099",
-"EMAIL","47","const char*",
+"EMAIL","50","const char*",
 "FEATURE","48","icalparameter_feature",X=22100;AUDIO;CHAT;FEED;MODERATOR;PHONE;SCREEN;VIDEO;NONE=22199",
 "LABEL","49","const char*",
-"EMAIL","50","const char*",
 "#VPATCH Extension Parameters","draft-daboo-icalendar-vpatch"
 "PATCH-ACTION","51","icalparameter_patchaction",X=22200;CREATE;BYNAME;BYVALUE;BYPARAM;NONE=22299",
 "#NOTE for updaters.  Preserve the icalparameter_kind Enum values to aid forward compatibility"


mcrha:
Some notes:
* I'm not sure how much that is an ABI change, but as the two were clashing it can be fine (for those not using `ICAL_REASON_PARAMETER`)
* an ideal state would be to update `scripts/readvaluesfile.pl` to check for both "key" duplicates and enum "value" duplicates and abort on clash, but I do not speak perl-lish, I'm sorry. Maybe someone knowledgeable can do it.

winterz:
For the 3.0 branch, I'm okay with this change except for moving changing the ICAL_REASON_PARAMETER enum from 43 to 47.  Unless there is some larger breakage caused by this?

In master I'm happy to apply the full patch and add some sanity checking to `scripts/readvaluesfile.pl`

winterz:
@mcrha @ksmurchison @ksuther look at design-data/properties.csv and you'll see 2 settings for a NAME property:

"NAME","51","TEXT","TEXT"
"NAME","115","TEXT","TEXT"

the second setting takes precedence in the generated code (i.e. `ICAL_NAME_PROPERTY = 115`) and the first setting is ignored.

The NAME property descriptions are:
* CAP Properties","RFC 4324 Section 2.1.2",   NAME -  The "NAME" property is used to add locale-specific descriptions into components.
*   Properties for iCalendar","RFC 7986 Section 5",  NAME - This property specifies the name of the calendar.

however, I don't see any reason to that we can't use the same set of getters/setters and enums, etc

IOW: correct me if I'm wrong, I think we can safely remove "NAME","51","TEXT","TEXT".   It isn't causing any code to be generated.

ksuther:
It looks like `"NAME","115","TEXT","TEXT"` was added later (7ca312debe69beafd6303ecc24c7581066a24303). `"NAME","51","TEXT","TEXT"` was added in 1ed39483cb05b92f80d070d2b46283e1494af12b. Do the integers have any meaning or are they just to provide a stable ordering? If that's all they do then certainly seems like the duplicate can go.

ksmurchison:
@winterz I think your analysis is correct.

As an side, do we even care about CAP?  Did/do any clients actually use it?

winterz:
> As an side, do we even care about CAP? Did/do any clients actually use it?
No idea. :)

winterz:
> Do the integers have any meaning or are they just to provide a stable ordering? If that's all they do then certainly seems like the duplicate can go.

Ideally they have a context with unique integers and setter/getters and the RFCs (IMO) shouldn't reuse the same identifier string.  In practice, for us I don't think it matters very much.  I'll comment the CAP NAME property

winterz:
@mcrha see master for commits:
 - e9c1b56aadc077a5316e2009e64863717faf4997  design-data/properties.csv - remove dupe entry for "NAME"
 - b92e1a41c59b4a829736580ddf73ac574344900e  design-data/value-types.csv - fix enum ICAL_ICAL_PROXIMITY_VALUE
 - 43cb2e215a8ed10e03989f19db5457f89fc2412a  design-data/parameters.csv: Fix enum value clashing (This patch)
 - 61ecac16f68c7c7c1459e1ebf708edb91c373586  scripts/readvaluesfile.pl - add some sanity checking
 
I'd rather not touch the 3.0 branch unless there is a really good reason.  
Else please close this pull request.

mcrha:
When the REASON and REQUIRED use the same value in the enum you can get either the first or the second when you search for the second or for the first. I mean:

* the component has set REASON, you search for ICAL_REQUIRED_PARAMETER, but libical returns REASON;
* and vice versa

because they use the same identifier in the enum.

I'm fine if you change this only in the master. I'm not using these parameters myself and the duplicate EMAIL is not causing any real trouble for the libical consumers.

> see master for commits:

nice, thanks.

Issue #531, opened 2021-12-12
No way to get the exact GEO value as text


dilyanpalauzov:
When `GEO:49.42612;7.75473` is parsed and converted with icalvalue_geo_as_ical_string_r() back to string, the result is `49.426120;7.754730`.  In the input the precision is 1.1m, in the output the precision is 11cm.

* Provide means to get the value`49.42612;7.75473` from the input.

winterz:
verified.  I'll see what I can do

winterz:
to solve this we could store GEO values internally as strings, not floats

ksmurchison:
That would make sense, as there is an iCalendar extension coming down the pike to allow geo: uris in GEO, e.g.: GEO;VALUE=URI:geo:37.786971,-122.399677;u=35
So we would probably store that as a string anyways

winterz:
Possible fix in PR https://github.com/libical/libical/pull/578
please review and test

dilyanpalauzov:
When the supplied test passes, I do not see a need to test further.

ksmurchison:
@winterz We recently had a client try to store a GEO property where one of the values was 3.035945999999967 and the property was rejected.  I'm wondering is we should increase ICAL_GEO_LEN, or since 8 decimal places has an accuracy of 1.1mm, we just truncate (and round?) these values after 8 decimal places.

Thoughts?

dilyanpalauzov:
<img src="https://imgs.xkcd.com/comics/coordinate_precision.png" alt="https://xkcd.com/2170/">

My understanding is that 5 digits after the decimal comma correspond to 1.1m precision - http://wiki.gis.com/wiki/index.php/Decimal_degrees - and no device can measure higher precision.

winterz:
> @winterz We recently had a client try to store a GEO property where one of the values was 3.035945999999967 and the property was rejected. I'm wondering is we should increase ICAL_GEO_LEN, or since 8 decimal places has an accuracy of 1.1mm, we just truncate (and round?) these values after 8 decimal places.
> 
> Thoughts?

I think truncate and round make sense.

ksmurchison:
@winterz  Another possible issue before I start on this.  Has anyone actually tried this code on Windows?  The documentation for GetNumberFormat() says that the input string must be NULL-terminated and that a leading '+' isn't allowed: https://learn.microsoft.com/en-us/windows/win32/api/winnls/nf-winnls-getnumberformata

[...]

PR #578, opened 2022-06-09
Internally represent GEO properties as text

winterz:
This allows arbitrary precision for the GEO values

fixes: #531


rsto:
Would this break compatibility with the Python wrapper? I see in design-data/value-types.cvs that it returns a tuple in Python, but is the current Python return value a tuple of floats, and now would become a tuple of strings?

Incidentally, I am currently also working on the GEO property. We plan to propose also allowing URI value types ([RFC 5870](https://www.rfc-editor.org/rfc/rfc5870.html)) for the `GEO` property, e.g.

    GEO;VALUE=URI:geo:37.786971,-122.399677;u=35

The geo: URI scheme allows for three-dimensional coordinates, uncertainty and a couple of additional improvements over the iCalendar GEO value. Apple is using them in the `X-APPLE-STRUCTURED-LOCATION` properties.

I think that this is independent of the change in PR, but I thought I share the news.

ksmurchison (approved):
LGTM

winterz:
> Would this break compatibility with the Python wrapper? I see in design-data/value-types.cvs that it returns a tuple in Python, but is the current Python return value a tuple of floats, and now would become a tuple of strings?

Yes it will.  I can fix.  Except this brings up even more issues:
 - the test.py is written in python2 => needs to be converted to python3
 - test.py should be invoked as part of the 'make test' target
 - actually, all of src/python should be ported to python3

> 
> Incidentally, I am currently also working on the GEO property. We plan to propose also allowing URI value types ([RFC 5870](https://www.rfc-editor.org/rfc/rfc5870.html)) for the `GEO` property, e.g.
> 
> ```
> GEO;VALUE=URI:geo:37.786971,-122.399677;u=35
> ```
> 
> The geo: URI scheme allows for three-dimensional coordinates, uncertainty and a couple of additional improvements over the iCalendar GEO value. Apple is using them in the `X-APPLE-STRUCTURED-LOCATION` properties.
> 
> I think that this is independent of the change in PR, but I thought I share the news.
And very good news indeed.  That's great!

winterz:
um.. side question.  are you actively using src/python?  I just realized we don't even build src/python 
maybe you're talking about the stuff in libical-glib?  If so, I handled that so the python API does not change.

winterz:
@rsto what's missing is the validation of the geo values (ie. latitiude <=abs(90.0) and longitude <= abs(180.) or whatever the standard says

maybe you can add that when working on the RFC 5870 stuff?

rsto:
> um.. side question. are you actively using src/python?

No, I just looked at the Python mapping for GEO when I learned how the values-types.csv file works.

> what's missing is the validation of the geo values

I'll make sure to validate these when I update the GEO handling code.

PR #580, opened 2022-06-14
libical-glib: Use the C locale for GEO string storage

tintou:
Make sure that we are never producing invalid ical by using locale-independant
conversions.


winterz:
merged in commit 813da5905f31cb7a62eecd3839bcaa02abe00d33

Issue #576, opened 2022-06-03
README.md


winterz:
look at the README at https://github.com/libical/libical
you'll see a couple strange things relating to markdown links.

1. "Libical — an implementation of iCalendar protocols and data formats {#mainpage}"
2. "For a conceptual overview of the library, see [UsingLibical.md](@ref UsingLibical)."  <- and there is no hyperlink created pointing to UsingLibical

winterz:
@mcclurgm   could you take a look?

mcclurgm:
Yes, this is an unfortunate side effect of merging the markdown files for use in Doxygen. The `{#mainpage}` label seems to be required to use a file that's not named `mainpage.md` as the Doxygen homepage. These artifacts will also appear wherever we want to use Doxygen header links in other files, too.

There's some claims on the internet that this style of header label is supported by GitHub Markdown, but that seems to be clearly wrong.

mcclurgm:
I should also note that I've tried to use standard Markdown links, which Doxygen says are supported, but fail on my system. If you link directly to `[UsingLibical](UsingLibical.md)` it fails to generate the link, although that would work properly in GitHub. I haven't figured out why, since the Doxygen documentation claims that it should work.

winterz:
not a big deal. thanks for looking.

winterz:
my Fedora system (where I typically generate the docs) has version 1.9.1
I'm happy to build doxygen 1.9.3 from source if necessary.

Here are the instructions I use when updating the dox => https://github.com/libical/libical/wiki/For-the-Maintainers%3A-Updating-the-API-Documentation-Pages.  I can easily add "make sure to use doxygen 1.9.3 or above".  BTW: do you have write access to that wiki?

winterz:
https://libical.github.io/libical/apidocs/index.html updated just now using doxygen1.9.3.

winterz:
all fixed.
thanks @mcclurgm

PR #581, opened 2022-06-15
Use plain Markdown link to UsingLibical.md in README.md

mcclurgm:
Fixes another issue in #576 

Seemingly requires Doxygen 1.9.3 to generate the link properly. I do not have 1.9.3 available on my system. Could you check to make sure that the link is working?

This is probably as far as we're going to get in terms of fixing Doxygen-specific Markdown artifacts. But with this change, at least README.md should render properly on GitHub.

### DIFF
--- a/README.md
+++ b/README.md
@@ -54,7 +54,7 @@ See the top-level [Install.txt](Install.txt) file.
 
 Documentation is hosted at <https://libical.github.io/libical/apidocs/>.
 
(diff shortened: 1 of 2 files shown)


winterz:
works locally with doxygen 1.9.3
I will merge and re-generate the online documentation

mcclurgm:
Thanks!

PR #582, opened 2022-06-19
Link RFCs to README using Doxygen link

### LINKED ISSUE #566: RFCS.md
Rather than duplicating the RFCs we support in several places (README.md, Mainpage.dox, UsingLibical.md) can we have one RFCS.md that contains the authoritative list and we embed that into the other md and dox files

CC: @mcclurgm

mcclurgm:
Fixes #566 

Following the suggestion there, move all RFC information to `README.md` and link from `UsingLibical.md` instead.

This version uses Doxygen header anchors, which are incompatible with GitHub. It will reintroduce artifacts to the GitHub-rendered version of the README, as previously seen in #576. Alternative solutions given in #566, which I could change to.

### DIFF
--- a/doc/UsingLibical.md
+++ b/doc/UsingLibical.md
@@ -11,23 +11,10 @@ and protocol data units. The iCalendar specification describes how
 calendar clients can communicate with calendar servers so users can
 store their calendar data and arrange meetings with other users.
 
-Libical implements [RFC5545][], [RFC5546][], [RFC7529][];
-the CalDav scheduling extensions in [RFC6638][];
-the iCalendar extensions in [RFC7986][], [RFC9073][], [RFC9074][];
-and some of [RFC6047][].
-
+Libical implements multiple [RFC calendar standards](@ref about).
 This documentation assumes that you are familiar with the iCalendar
 standards RFC5545 and RFC5546. These specifications are available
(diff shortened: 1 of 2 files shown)


doc/UsingLibical.md
```diff
@@ -11,23 +11,10 @@ and protocol data units. The iCalendar specification describes how
 calendar clients can communicate with calendar servers so users can
 store their calendar data and arrange meetings with other users.
 
-Libical implements [RFC5545][], [RFC5546][], [RFC7529][];
-the CalDav scheduling extensions in [RFC6638][];
-the iCalendar extensions in [RFC7986][], [RFC9073][], [RFC9074][];
-and some of [RFC6047][].
-
+Libical implements multiple [RFC calendar standards](@ref about).
```
mcclurgm, line 14: An alternative would look like this:

```suggestion
Libical implements multiple RFC calendar standards, listed in the [README](../README.md).
```

We could then remove the `{#about}` tag from the README.

winterz: Thanks .
I took the first idea and modified it a bit by add a new "RFCs" section to the README.me

new version is up at http://libical.github.io/libical/apidocs/index.html

Issue #589, opened 2022-09-14
icaltime_as_timet yields incorrect results after year 2100


minichma:
icaltime_as_timet yields incorrect results for dates/times at or after 2100-03-01.

Consider the following test case, which reproduces the issue (`ok` macro as used in `regresion.c`). The last date/time that yields a correct result is `21000228T235959Z`.

```C
void test_icaltime_as_timet(void) {

    ok("icaltime_from_string translates 19000101T000000Z to -2209075200", icaltime_as_timet(icaltime_from_string("19000101T000000Z")) == -2209075200);
    ok("icaltime_from_string translates 19700101T000000Z to 0", icaltime_as_timet(icaltime_from_string("19700101T000000Z")) == 0);
    ok("icaltime_from_string translates 19700301T235959Z to 5183999", icaltime_as_timet(icaltime_from_string("19700301T235959Z")) == 5183999);
    ok("icaltime_from_string translates 19970717T235959Z to 869183999", icaltime_as_timet(icaltime_from_string("19970717T235959Z")) == 869183999);
    ok("icaltime_from_string translates 20241202T235959Z to 1733183999", icaltime_as_timet(icaltime_from_string("20241202T235959Z")) == 1733183999);
    ok("icaltime_from_string translates 20371231T235959Z to 2145916799", icaltime_as_timet(icaltime_from_string("20371231T235959Z")) == 2145916799);
#if (SIZEOF_TIME_T > 4)
    ok("icaltime_from_string translates 20520419T235959Z to 2597183999", icaltime_as_timet(icaltime_from_string("20520419T235959Z")) == 2597183999);
    ok("icaltime_from_string translates 20790905T235959Z to 3461183999", icaltime_as_timet(icaltime_from_string("20790905T235959Z")) == 3461183999);
    ok("icaltime_from_string translates 21000101T235959Z to 4102531199", icaltime_as_timet(icaltime_from_string("21000101T235959Z")) == 4102531199);
    ok("icaltime_from_string translates 21000228T235959Z to 4107542399", icaltime_as_timet(icaltime_from_string("21000228T235959Z")) == 4107542399);

// Fails from here on
    ok("icaltime_from_string translates 21000301T000000Z to 4107542400", icaltime_as_timet(icaltime_from_string("21000301T000000Z")) == 4107542400);
    ok("icaltime_from_string translates 22370122T235959Z to 8427628799", icaltime_as_timet(icaltime_from_string("22370122T235959Z")) == 8427628799);
[...]

PR #590, opened 2022-09-15
Fix icaltime_as_timet to support dates after year 2100.

minichma:
See #589 - icaltime_as_timet used to return inaccurate results for dates starting with 2100-03-01. With this PR the function is fixed to support dates up to year 10000. Values before 1902-01-01 are still inaccurate. Even 19011213T204552Z, which is within the 32-bit range, doesn't work properly. The defined range is documented in the function's header doc.

### DIFF
--- a/src/libical/icaltime.c
+++ b/src/libical/icaltime.c
@@ -102,6 +102,7 @@ static icaltime_t icaltime_timegm(const struct tm *tm)
 static icaltime_t make_time(struct tm *tm, int tzm)
 {
     icaltime_t tim;
+    int febs;
 
     static int days[] = { -1, 30, 58, 89, 119, 150, 180, 211, 242, 272, 303, 333, 364 };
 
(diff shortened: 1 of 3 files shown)


minichma:
@winterz The build fails on Ubuntu but the issue doesn't seem to be related to this PR. How do we deal with this?

winterz:
> @winterz The build fails on Ubuntu but the issue doesn't seem to be related to this PR. How do we deal with this?

the failures are something strange in the doxygen package.  not related to your PR.  I resubmitted the build because looks like an AppVeyor hiccup.

I notice this patch is for the master branch.  would it also make sense in the 3.0 branch?

winterz:
https://github.com/appveyor/ci/issues/3842

winterz:
can you rebase for e254ef8f8797b0ad30cceeeade869fbcf1e70264

minichma:
rebased, but now it conflicts with master; or did you mean just to create an additional branch/PR for 3.0?

minichma:
Rebased this PR on top of master and created #591, back-porting the change to 3.0.

winterz:
merged 6e4dab56cf1383b4d2650be019199eedf4f22755 from the 3.0 branch

then fixed the SIZEOF_TIME_T => SIZEOF_ICALTIME_T change in commit 4e95d6152a8ebffdcc1eb2b960b41f6280fe455c

PR #591, opened 2022-09-16
Fix icaltime_as_timet to support dates after year 2100.

minichma:
This is a back-port of #590 to the v3.0 baseline.

### DIFF
--- a/src/libical/icaltime.c
+++ b/src/libical/icaltime.c
@@ -111,6 +111,7 @@ static time_t icaltime_timegm(const struct tm *tm)
 static time_t make_time(struct tm *tm, int tzm)
 {
     time_t tim;
+    int febs;
 
     static int days[] = { -1, 30, 58, 89, 119, 150, 180, 211, 242, 272, 303, 333, 364 };
 
@@ -145,12 +146,17 @@ static time_t make_time(struct tm *tm, int tzm)
 
     tim = (time_t) ((tm->tm_year - 70) * 365 + ((tm->tm_year - 1) / 4) - 17);
 
+    /* adjust: no leap days every 100 years, except every 400 years. */
+
+    febs = (tm->tm_year - 100) - ((tm->tm_mon <= 1) ? 1 : 0);
+    tim -= febs / 100;
+    tim += febs / 400;
+
     /* add number of days elapsed in the current year */
 
     tim += days[tm->tm_mon];
 
-    /* check and adjust for leap years (the leap year check only valid
-       during the 32-bit era */
+    /* check and adjust for leap years */
 
     if ((tm->tm_year & 3) == 0 && tm->tm_mon > 1)
         tim += 1;
--- a/src/test/regression.c
+++ b/src/test/regression.c
@@ -4945,6 +4945,29 @@ static void test_remove_tzid_from_due(void)
     icalcomponent_free(c);
 }
 
+void test_icaltime_as_timet(void) {
+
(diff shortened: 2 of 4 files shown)


src/libical/icaltime.c
```diff
@@ -145,12 +145,17 @@ static time_t make_time(struct tm *tm, int tzm)
 
     tim = (time_t) ((tm->tm_year - 70) * 365 + ((tm->tm_year - 1) / 4) - 17);
 
+    /* adjust: no leap days every 100 years, except every 400 years. */
+
+    int febs = (tm->tm_year - 100) - ((tm->tm_mon <= 1) ? 1 : 0);
```
winterz, line 150: need to declare 'int febs' near the top of the function, to support ancient C

src/test/regression.c
```diff
@@ -4945,6 +4945,29 @@ static void test_remove_tzid_from_due(void)
...
+    ok("icaltime_from_string translates 19290519T000000Z to -1281916800", icaltime_as_timet(icaltime_from_string("19290519T000000Z")) == -1281916800);
+    ok("icaltime_from_string translates 19561004T000000Z to -417916800", icaltime_as_timet(icaltime_from_string("19561004T000000Z")) == -417916800);
+    ok("icaltime_from_string translates 19700101T000000Z to 0", icaltime_as_timet(icaltime_from_string("19700101T000000Z")) == 0);
+    ok("icaltime_from_string translates 19700301T235959Z to 5183999", icaltime_as_timet(icaltime_from_string("19700301T235959Z")) == 5183999);
+    ok("icaltime_from_string translates 19970717T235959Z to 869183999", icaltime_as_timet(icaltime_from_string("19970717T235959Z")) == 869183999);
+    ok("icaltime_from_string translates 20241202T235959Z to 1733183999", icaltime_as_timet(icaltime_from_string("20241202T235959Z")) == 1733183999);
+    ok("icaltime_from_string translates 20371231T235959Z to 2145916799", icaltime_as_timet(icaltime_from_string("20371231T235959Z")) == 2145916799);
+#if (SIZEOF_ICALTIME_T > 4)
```
winterz, line 4958: the 3.0 branch doesn't have a macro for SIZEOF_ICALTIME_T 
I guess you'll need to add:
 * `check_type_size(${ICAL_ICALTIME_T_TYPE} SIZEOF_ICALTIME_T)` to ConfigureChecks.cmake
 * `#cmakedefine SIZEOF_ICALTIME_T ${SIZEOF_ICALTIME_T}` to config.h.cmake

winterz:
also please add a blurb about this into the 3.0.15 section of ReleaseNotes.txt

src/libical/icaltime.c
minichma: Changed as suggested. Although this should be supported by C99, which is the minimum we require.

winterz: you're right.  I'm using a tool (splint) to check for ancient C support.  I'll see if I can tell splint to allow C99

src/test/regression.c
minichma: Thanks! We don't have icaltime_t in 3.0. We're still using time_t, so we can simply check SIZEOF_TIME_T. Fixed it.

minichma:
Added comment to ReleaseNotes.txt. How should I deal with the ReleaseNotes on master? Anything to do there?

winterz:
committed in 6e4dab56cf1383b4d2650be019199eedf4f22755

PR #593, opened 2022-09-23
Fixes to x-property comma handling and escaping

### LINKED ISSUE #592: Should x-properties allow multiple values?
Apple uses unescaped commas in their x-property values. For example:

```
X-APPLE-TRAVEL-START;VALUE=URI:geo:99.000000,-99.000000
X-CALENDARSERVER-ATTENDEE-COMMENT;X-CALENDARSERVER-ATTENDEE-REF="urn:x-uid
 :XXXXXXXX-XXXX-XXXX-XXXX-XXXXXXXXXXXX";X-CALENDARSERVER-DTSTAMP=20220101T
 000000Z:Test, test.
```

Reading RFC 5545, my interpretation is that we shouldn't be splitting these into multiple values during parsing.

https://www.rfc-editor.org/rfc/rfc5545#section-3.1.2:
> ... However, it should be noted that some properties
   support encoding multiple values in a single property by separating
   the values with a COMMA character.  Individual property definitions
   should be consulted for determining whether a specific property
   allows multiple values and in which of these two forms. ...

The way I read it, only properties that explicitly say they can contain multiple values should be split.

If that's the case, x-properties should be ignored in `parser_get_next_value()` and `icalmemory_strdup_and_quote()`. Any opinions (@winterz @ksmurchison)?

ksuther:
- Don't allow x-properties to be split into multiple values since RFC 5545 says that only explicitly documented properties should be split
- Don't escape text in x-property values since RFC 5545 says that the value should be *VALUE-CHAR rather than *SAFE-CHAR

These changes allow Apple's x-properties (such as X-CALENDARSERVER-ATTENDEE-COMMENT and X-APPLE-TRAVEL-START) to be round-tripped safely.

Fixes #592.

### DIFF
--- a/src/libical/icalparser.c
+++ b/src/libical/icalparser.c
@@ -409,8 +409,8 @@ static char *parser_get_next_value(char *line, char **end, icalvalue_kind kind)
                 continue;
             }
         }
-        /* ignore all , for query value. select dtstart, dtend etc ... */
-        else if (kind == ICAL_QUERY_VALUE) {
+        /* ignore all commas for query and x values. select dtstart, dtend etc ... */
+        else if (kind == ICAL_QUERY_VALUE || kind == ICAL_X_VALUE) {
             if (next != 0) {
                 p = next + 1;
                 continue;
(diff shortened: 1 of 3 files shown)


ksmurchison (approved):
LGTM

winterz:
add a blurb about this into the ReleaseNotes.txt please.

winterz:
why does the commit message look like one big run-on sentence for me?

```    Fixes to x-property comma handling and escaping - Don't allow x-properties to be split into multiple values since RFC 5545 says that only explicitly documented properties should be split - Don't escape text in x-property values since RFC 5545 says that the value should be *VALUE-CHAR rather than *SAFE-CHAR These changes allow Apple's x-properties (such as X-CALENDARSERVER-ATTENDEE-COMMENT and X-APPLE-TRAVEL-START) to be round-tripped safely```

winterz:
I cleaned-up the commit message and added the blurb to ReleaseNotes.txt

PR #607, opened 2022-10-10
Fix argument guards in icaltime_as_timet to match documentation and tests.

minichma:
With 3.0.15 we introduced tests for `icaltime_as_timet`, checking for values starting with 1902. With 64-bit time_t this works just fine, but with 32-bit version the tests fail due to a more strict validation of input parameters that doesn't allow for dates < year 1970. With this PR we change the argument validation of `icaltime_as_timet`. We now allow dates starting with 1902 also for 32-bit time_t and we also validate the date for 64-bit time_t, allowing for 1902-01-01 through 10000-01-01 (excl.).

### DIFF
--- a/ReleaseNotes.txt
+++ b/ReleaseNotes.txt
@@ -4,6 +4,7 @@ Release Highlights
 Version 3.0.16 (UNRELEASED):
 ----------------------------
  * Fix a regression in 3.0.15 that limited how many lines could be processed in one call to icalparser_parse()
+ * Fix argument guards in icaltime_as_timet to match documentation and tests.
 
 Version 3.0.15 (06 October 2022):
 ---------------------------------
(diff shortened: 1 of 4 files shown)


minichma:
@winterz I also added more tests in regression.c, now testing every day between 1902 and year 10000, which is quite a lot. The tests take only ~250 ms on my moderately powered laptop, so I think this should be ok. If you think we should reduce this, I'm happy to do so.

I also prepared a merge branch for master but no PR so far: https://github.com/libical/libical/tree/fix_timet_32bit_1902_master

winterz:
@minichma I'm going to commit a separate fix for all those pre-commit whitespace issues.  then please rebase this

winterz:
ready. hopefully all the pre-commit issues are fixed now

winterz:
merged

PR #609, opened 2022-10-11
Concurrency mode thread-local

minichma:
Today it is not easily possible to use multiple instances of libical in parallel, mainly due to the state being stored in global variables. With pthreads we have a way of synchronizing access, so the internal state doesn't get corrupted in case of concurrent access. But on a higher level additional synchronization is still required, because individual threads might modify the global state independently.

With this PR libical can be built in the new 'thread-local' concurrency mode. If enabled, all global variables are marked as `thread_local` (or the compiler-specific equivalent). This way any code accessing libical can do so without additional synchronization, which makes access easier and faster. However, initialization of the library needs to be done for each thread.

The functionality may especially be helpful when writing high-level wrappers. In future work this might be a good mode for implementing the GLIB or Java wrappers. It can also be helpful to allow unit tests to run in parallel.

The feature can be enabled via the `LIBICAL_SYNCMODE_THREADLOCAL=[true|false]` CMAKE option. It's marked as experimental for now, so we can make adjustments if helpful. It would be better to have something like `LIBICAL_SYNCMODE=[none|pthread|threadlocal]` but I'm not sufficiently into CMAKE to implement this quickly. Help would be appreciated.

Could also be back-ported to 3.1 if deemed helpful.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -107,6 +107,13 @@
 #  Build with the undefined sanitizer (requires gcc or clang)
 #  Default=false
 #
+# -DLIBICAL_SYNCMODE_THREADLOCAL=[true|false]
+#  Experimental: If set to true, global variables are marked as thread-local. This allows accessing
+#  libical from multiple threads without the need for synchronization. However, any global settings
+#  must be applied per thread.
+#  If set to false, the default synchronization mechanism is applied. That is, if the pthreads library
+#  is available, it will be used, otherwise no synchronization is applied.
+#  Default: false.
 
 cmake_minimum_required(VERSION 3.11.0) #first line, to shutup a cygwin warning
 project(libical C) #CXX is optional for the bindings
@@ -735,6 +742,9 @@ if(LIBICAL_DEVMODE_UNDEFINED_SANITIZER)
     set(ICAL_GLIB False)
   endif()
 endif()
+
+mark_as_advanced(LIBICAL_SYNCMODE_THREADLOCAL)
+libical_option(LIBICAL_SYNCMODE_THREADLOCAL "Experimental: Mark global variables as thread-local." False)
 
 libical_option(ENABLE_LTO_BUILD "Build a link-time optimized version." False)
 if(ENABLE_LTO_BUILD)
--- a/config.h.cmake
+++ b/config.h.cmake
@@ -582,3 +582,6 @@ typedef ssize_t IO_SSIZE_T;
 #define ICALMEMORY_DEFAULT_MALLOC malloc
 #define ICALMEMORY_DEFAULT_REALLOC realloc
 #define ICALMEMORY_DEFAULT_FREE free
+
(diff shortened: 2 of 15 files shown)


winterz:
the appveyor failure is a CI problem; not the fault of this patch.

@minichma I did some local testing and found the builtin_timezones test fails:
```builtin_timezones: /home/allen/projects/libical/libical/src/test/builtin_timezones.c:45: thread_func: Assertion `thread_comp == icalcomp' failed.```

**Comment by minichma** [comment]
@winterz I fixed the builtin_timezones test for the threadlocal configuration on linux. Sorry, missed this case, as the problem doesn't appear on Windows, where there's no pthread.

This whole PR is more or less a suggestion to be discussed. It's an easy solution for the concurrency challenges that come with the global vars. A better approach would certainly be to move all the globals to some context object and pass it through to all the places where they are needed (which will be close to everywhere), but this would be a major change of the API could therefore only be considered for 4.0, while the suggested change is fully compatible. Please let me know what you think!

**Comment by winterz** [comment]
I'm ok with this approach.

I do get a SEGFAULT in icaltm_test however when compiled on linux+gcc with `cmake -DCMAKE_BUILD_TYPE=Debug`

```
Program terminated with signal SIGSEGV, Segmentation fault.
#0  0x00007fe9a81df0f8 in icalarray_element_at (array=0x7fe998003f10, position=position@entry=18446744073709551615) at /home/allen/projects/libical/libical/src/libical/icalarray.c:127
127         return (char *)(array->chunks[chunk]) + (offset * array->element_size);
[Current thread is 1 (Thread 0x7fe9a55fe640 (LWP 690146))]
#0  0x00007fe9a81df0f8 in icalarray_element_at (array=0x7fe998003f10, position=position@entry=18446744073709551615) at /home/allen/projects/libical/libical/src/libical/icalarray.c:127
#1  0x00007fe9a81f464d in icaltimezone_get_utc_offset (zone=zone@entry=0x139da28, tt=tt@entry=0x7fe9a55fbf70, is_daylight=is_daylight@entry=0x0) at /home/allen/projects/libical/libical/src/libical/icaltimezone.c:922
#2  0x00007fe9a81f4a0f in icaltimezone_convert_time (tt=tt@entry=0x7fe9a55fbf70, from_zone=from_zone@entry=0x139da28, to_zone=to_zone@entry=0x7fe9a873c7e8) at /home/allen/projects/libical/libical/src/libical/icaltimezone.c:800
#3  0x00007fe9a81ef986 in icaltime_convert_to_zone (tt=..., zone=0x7fe9a873c7e8) at /home/allen/projects/libical/libical/src/libical/icaltime.c:844
```

minichma:
Fixed icaltm_test, thanks for pointing this out. All tests I'm able to run should work for Linux and Windows, both for thread-local mode enabled and disabled. However, I'm not able to run the glib tests here, so I'm not sure about those. But probably it doesn't make too much sense to build libical-glib in thread-local mode anyways. I guess, there would be more to think about.

winterz:
this patch has been committed

PR #619, opened 2022-10-24
disable .so lib versions for Android

chenhx37:


winterz:
committed into the 3.0 branch with d59d183c62df82d4e8db709720478214b9c28805
then merged into master

I also removed the ancient android toolchain file and added some installation help to Install.txt

thanks for your contribution

PR #623, opened 2022-10-26
misc: Update sources to more modern C

mcrha:
The code fails to build when having enabled (gcc) warnings:
   -Werror=implicit
   -Werror=implicit-function-declaration
   -Werror=implicit-int
   -Werror=int-conversion
   -Werror=old-style-definition
   -Werror=strict-prototypes

### DIFF
--- a/examples/access_components.c
+++ b/examples/access_components.c
@@ -24,7 +24,7 @@ void do_something(icalcomponent *c);
 
 */
 
-icalcomponent* create_new_component()
+icalcomponent* create_new_component(void)
 {
 
     /* variable definitions */
@@ -193,7 +193,7 @@ icalcomponent* create_new_component()
 /* Now, create the same component as in the previous routine, but use
 the constructor style. */
 
-icalcomponent* create_new_component_with_va_args()
+icalcomponent* create_new_component_with_va_args(void)
 {
 
     /* This is a similar set up to the last routine */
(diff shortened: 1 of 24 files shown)


winterz:
non-modern stuff will sneak back in unless we add those compile options by default.

in the top-level CMakeLists.txt maybe we need to add
```
set(CMAKE_C_FLAGS "${CMAKE_C_FLAGS} -Werror=implicit -Werror=implicit-function-declaration -Werror=implicit-int -Werror=int-conversion -Werror=old-style-definition -Werror=strict-prototypes")
```
I also didn't want to force "modern C" into the 3.0 branch.  this would be fine in the master branch though.
Do we need this in 3.0?

mcrha:
I can modify the CMakeLists.txt. Not sure about the -Werror... thing, maybe just enable the warning is enough.

With respect of the 3.0, I've got an impression that it's the branch for the current development. It's not needed for the 3.0 at the moment and it won't be needed for some time. The idea behind this is from the Fedora's proposed change: https://lists.fedoraproject.org/archives/list/devel@lists.fedoraproject.org/thread/CJXKTLXJUPZ4F2C2VQOTNMEA5JAUPMBD/
 I might mention it earlier, I'm sorry.

winterz:
appveyor CI is showing compile errors

mcrha:
> /home/appveyor/projects/libical/src/test/icaltm_test.c:30:25: error: parameter name omitted

aha, they do not like unnamed parameters. It's fixed now.

a17r:
This should fix build with >=clang-16 (error in caldate.c), see also: https://bugs.gentoo.org/880391

thesamesam:
Thanks!

PR #584, opened 2022-08-24
vCard API

ksmurchison:
Add new API for constructing/serializing vCard data, similar to the API for iCalendar data.

PR #627, opened 2022-11-15
Make vCard parser reentrant

vtronko:
# Problem

Parsing multiple vCard documents in parallel can cause trouble, because parser is not reentrant.
>[Subroutine](https://en.wikipedia.org/wiki/Subroutine) is called reentrant if multiple invocations can safely run concurrently on multiple processors, or on a single processor system, where a reentrant procedure can be interrupted in the middle of its execution and then safely be called again ("re-entered") before its previous invocations complete execution.

# Issue reproduction

Here's a program that demonstrates the perils of current parser:
```
#include <iostream>
#include <fstream>
#include <sstream>
#include <future>
#include <vector>
#include <mutex>

#include <libical/ical.h>
#include <libical/vcc.h>
#include <libical/vobject.h>

int main() {
  std::string content;
  {
    std::ifstream myfile("test.vcf");
    std::stringstream ss;
    ss << myfile.rdbuf();
    content = ss.str();
  }

  using vcardptr = std::unique_ptr<VObject, decltype(&deleteVObject)>;
  std::vector<std::future<vcardptr>> results;
  for (int i = 0; i < 100; ++i) {
    std::string src = content;
    results.push_back(std::async(std::launch::async, [=] {
        // comment these 2 lines to crash
        /* static std::mutex mut; */
        /* std::unique_lock<std::mutex> _ { mut }; */

        vcardptr ptr(Parse_MIME((char*)src.c_str(), src.size()), deleteVObject);
        return ptr;
    }));
  }
  
  for (auto& f : results) {
    f.get();
  }
  
  return 0;
}
```

Once the ad-hoc synchronization is commented, all you get are `segmentation fault (core dumped) `

# Solution

Previously the parser code was generated via `byacc -p "mime_" -o vcc.c vcc.y`. It makes sense to use additional flag as following: `byacc -P -p "mime_" -o vcc.c vcc.y`

From man:
```
-P   create a reentrant parser, e.g., “%pure-parser”.
```
It can make sense to add `%pure-parser` in .y file instead, to avoid confusion with invocation params in future. It is up for debate which one is better for maintainability, this can always be adjusted.

Also, custom structures (like LexBuf) should use thread local storage. It is possible to solve this problem via function signatures, so pointer to lexBuf could be passed between calls, but in my opinion, `__thread` is more elegant.


minichma:
Related issue treated in https://github.com/libical/libical/pull/609.

minichma:
The SegFault you're seeing in the verification build could be related to memory management. In particular this happens if you free memory using stdlib functions (i.e. free() et al), which has been allocated using icalmemory_new_buffer() et al. The verification is enabled if you compile using the CMAKE switch `-DLIBICAL_DEVMODE_MEMORY_CONSISTENCY=true`, which is the case for the test cases.

vtronko:
@minichma I played with your branch locally, and tried to adopt the idea of `ICAL_GLOBAL_VAR` define wrapper. One place it didn't work for me though was `vobject.h`, where I'd like to have `extern ICAL_GLOBAL_VAR const char **fieldedProp;` instead of just `__thread`, and it would fail for user of the library because he understandably doesn't have access to `ICAL_GLOBAL_VAR` define. Any idea how I could circumvent that?

minichma:
@vtronko I don't know much about this code. Is `fieldedProp` considered public? In other words, would removing it from the header file be considered a breaking change? Doesn't look to me like that and I think, exposing global vars via the public APIs should be avoided anyways. If it can be considered private, it could be replaced by accessor functions. Or the declaration could be moved to some private header file. I guess, @winterz would be the one to comment on that.

winterz:
we can break source and binary compatibility in the master branch.
feel free

vtronko:
I am checking what happened to regression tests.

vtronko:
Unfortunately, I can't figure why exactly it happens. .y files are nightmare to debug.

vtronko:
Ahem, I figured it out, was a rookie mistake, some weird code made its way to the changeset.
Ran tests, they all work. So, waiting for CI and I guess it can be reviewed.

ksmurchison:
Personally, I think the existing vCard parser is lacking and we should focus on #584 (or something like it).

winterz:
this patch introduces memory leak

```
==460493==ERROR: LeakSanitizer: detected memory leaks

Direct leak of 10 byte(s) in 1 object(s) allocated from:
    #0 0x4d6f87 in malloc (/home/allen/projects/libical/libical/build-test1asan-asan/src/test/regression+0x4d6f87) (BuildId: 8b54a0d6be46ab45420cd240a8634b1e3b181c23)
    #1 0x7fbbebde09b4 in dupStr /home/allen/projects/libical/libical/src/libicalvcal/vobject.c:102:16
    #2 0x7fbbebdec9ba in lexStr /home/allen/projects/libical/libical/build-test1asan-asan/src/libicalvcal/vcc.y:617:12
    #3 0x7fbbebde8f79 in lexGetWord /home/allen/projects/libical/libical/build-test1asan-asan/src/libicalvcal/vcc.y:639:12
    #4 0x7fbbebde79b1 in mime_lex /home/allen/projects/libical/libical/build-test1asan-asan/src/libicalvcal/vcc.y:1069:35
    #5 0x7fbbebde98fa in mime_parse /home/allen/projects/libical/libical/build-test1asan-asan/src/libicalvcal/vcc.c:1472:23
    #6 0x7fbbebde90db in Parse_MIMEHelper /home/allen/projects/libical/libical/build-test1asan-asan/src/libicalvcal/vcc.y:1107:9
    #7 0x7fbbebde9085 in Parse_MIME /home/allen/projects/libical/libical/build-test1asan-asan/src/libicalvcal/vcc.y:1119:12
    #8 0x54df9c in test_vcc_vcard_parse /home/allen/projects/libical/libical/src/test/regression.c:5062:12
    #9 0x557f66 in test_run /home/allen/projects/libical/libical/src/test/regression-utils.c:236:9
    #10 0x54d390 in main /home/allen/projects/libical/libical/src/test/regression.c:5502:5
    #11 0x7fbbeae2950f in __libc_start_call_main (/lib64/libc.so.6+0x2950f) (BuildId: 85c438f4ff93e21675ff174371c9c583dca00b2c)

SUMMARY: AddressSanitizer: 10 byte(s) leaked in 1 allocation(s).`
```

winterz:
> Personally, I think the existing vCard parser is lacking and we should focus on #584 (or something like it).

ideally I agree.
I don't think it hurts to commit this patch (once the memory leak is fixed) in the meantime.

the code is so ancient I'll be very happy once we replace it

mcrha:
I would also set the `->str` to NULL after the deletion, to ensure no use-after-free is introduced.

Otherwise I agree, the error path handling does not free everything as it should.

mcrha:
I've been thinking of something similar what they suggest in the link you provided, to track the memory. The `dupStr()` and `deleteStr()` are defined functions in `vobject.c`, thus they can use some hash table (for performance reasons) of the allocated memory and free all remaining in the `finiLex()`.

Hash tables in C are tricky; glib has it implemented, but it's glib. I noticed you have there some hashing on strings already involved, with `strTbl` in the `vobject.c`. The new has table could use just the lower bytes of the address, not the string itself, as a hash value. I'm only thinking aloud.

PR #636, opened 2023-03-23
libical-glib: Add get_zone_directory

jtojnar:
This will be useful for evolution-data-server to use the same zone directory as libical.


winterz:
do you need this in the 3.0 branch as well?
as the master branch is a long way from being released and there's a good possibility a new 3.0.17 release will happen.

if so, please move this PR to the 3.0 branch.
also add an entry to the ReleaseNotes.txt about a new publicly available function called get_zone_directory

PR #638, opened 2023-03-29
Basic validation + ignore leading spaces when parsing .ics string

guidedways:
- Ignores leading spaces / line-breaks when fed poorly generated .ics files (some automated tools add spaces / leading line breaks)
- Performs basic validation before trying to parse a potentially malformed .ics string

### DIFF
--- a/src/libical/icalparser.c
+++ b/src/libical/icalparser.c
@@ -1335,8 +1335,14 @@ char *icalparser_string_line_generator(char *out, size_t buf_size, void *d)
 
 icalcomponent *icalparser_parse_string(const char *str)
 {
+    icalcomponent *c;
+    struct slg_data d;
+    icalparser *p;
+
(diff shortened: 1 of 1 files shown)


kdabci:
please add some regressions tests with that would trip this new validation code

src/libical/icalparser.c
```diff
@@ -1341,8 +1341,30 @@ icalcomponent *icalparser_parse_string(const char *str)
...
+        strStart++;
+    }
+
+    // All spaces?
+    if (*strStart == 0) {
+        goto error;
+    }
+
+    // basic validation
+    size_t beginSize = strlen("BEGIN");
```
dilyanpalauzov, line 1360: Please replace with a constant (5).  This way runtime overhead is avoided.

winterz:
@guidedways did you abandon this?

I think its a good idea.  but I would like to see some tests added to regression.c

PR #647, opened 2023-09-19
gitignore: ignore build/ directory rather than /build

rsto:
The `.gitignore` file ignores the `/build` directory, outside the source controlled file tree. This fixes it to actually ignore the `build/` directory local to the source directory.


ksmurchison:
This looks correct to me.  @winterz do you agree?

PR #651, opened 2023-10-10
Add icalparser_ctrl to handle CONTROL characters during parse

rsto:
RFC 5545 section 3.1 defines CONTROL characters to be byte values in the decimal ranges [0-8], [10-31] as well as 127. Of these, all but 10 (LF) and 13 (CR) are not allowed to be part of any valid content-line produced by the
"contentline" ABNF defined in this section.

However, the current libical parser allows such illegal values to occur at almost all places within iCalendar and preserves them in TEXT values as well as non-standard property names. This causes libical to silently accept invalid iCalendar input and generate invalid output for such data.

This patch adds the icalparser_ctrl setting to define how to handle invalid CONTROL characters during parse:

* KEEP: preserve current libical behaviour
* OMIT: silently ignores such characters during parse
* ERROR: rejects the content-line and inserts an error

For backwards-compability, the default setting is KEEP.

### DIFF
--- a/src/test/icalparser_ctrl_test.c
+++ b/src/test/icalparser_ctrl_test.c
@@ -1,0 +1,81 @@
+/*======================================================================
+ FILE: icalparser_ctrl_test.c
+ CREATOR: Robert Stepanek, 10 October 2023
+
+ SPDX-FileCopyrightText: 2000 Eric Busboom <eric@civicknowledge.com>
+
+ SPDX-License-Identifier: LGPL-2.1-only OR MPL-2.0
+
+======================================================================*/
+
+/*
+ * Program to test handling of CONTROL characters in iCalendar parsing.
+ */
+
+#ifdef HAVE_CONFIG_H
+#include <config.h>
+#endif
+
+#include "libical/ical.h"
+
+static void assert_ctrl(enum icalparser_ctrl ctrl,
+                        const char *data,
+                        const char *want_desc,
+                        int want_xlicerror)
+{
+    icalparser_set_ctrl(ctrl);
+
+    icalcomponent *ical = icalparser_parse_string(data);
+    assert(ical);
+
+    icalcomponent *comp = icalcomponent_get_first_real_component(ical);
+    assert(comp);
+
+    if (want_desc) {
+        assert(!strcmp(want_desc, icalcomponent_get_description(comp)));
+    } else {
+        assert(!icalcomponent_get_first_property(comp, ICAL_DESCRIPTION_PROPERTY));
+    }
+
+    if (want_xlicerror) {
+        assert(icalcomponent_get_first_property(comp, ICAL_XLICERROR_PROPERTY));
+    } else {
+        assert(!icalcomponent_get_first_property(comp, ICAL_XLICERROR_PROPERTY));
+    }
+
+    icalcomponent_free(ical);
+}
+
+int main(int argc, char *argv[])
+{
+    if (argc != 1) {
+        fprintf(stderr, "Usage: %s\n", argv[0]);
+        return 0;
+    }
+
+    // Assert default value
+    assert(icalparser_get_ctrl() == ICALPARSER_CTRL_KEEP);
+
+    // Assert icalparser_ctrl settings
+    static const char *data = ""
+        "BEGIN:VCALENDAR\r\n"
+        "VERSION:2.0\r\n"
+        "PRODID:-//ACME/DesktopCalendar//E\r\n"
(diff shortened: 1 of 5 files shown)


ksmurchison (approved):
I like it.  This solves a problem that we have recently seen in the wild.

winterz:
very nice indeed.
obviously fix the pre-commit and build issues first.
please also add a blurb into the ReleaseNotes.txt

another question: should we have this in the 3.0 version too?  I think we should.
I'd be willing to make a new 3.0.x release

ksmurchison:
> another question: should we have this in the 3.0 version too? I think we should. I'd be willing to make a new 3.0.x release

Yes, I think we should.

winterz:
then I will cherry-pick this into the 3.0 branch once it lands.  and make a next 3.0.x release

rsto:
I just removed the call to `basename` to make the macOS and Windows builds happy.

rsto:
And now updated the ReleaseNotes

src/test/icalparser_ctrl_test.c
```diff
@@ -0,0 +1,67 @@
+#ifdef HAVE_CONFIG_H
```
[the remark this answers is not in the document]
rsto: done

src/test/icalparser_ctrl_test.c
```diff
@@ -0,0 +1,81 @@
...
+                        const char *data,
+                        const char *want_desc,
+                        int want_xlicerror)
+{
+    icalparser_set_ctrl(ctrl);
+
+    icalcomponent *ical = icalparser_parse_string(data);
+    assert(ical);
+
+    icalcomponent *comp = icalcomponent_get_first_real_component(ical);
```
dilyanpalauzov, line 31: This produces a compiler warning:
```
[50/57] Building C object src/test/CMakeFiles/parser_ctrl.dir/icalparser_ctrl_test.c.o                                                  
/git/libical/src/test/icalparser_ctrl_test.c: In function ‘assert_ctrl’:                                                                
/git/libical/src/test/icalparser_ctrl_test.c:31:20: warning: unused variable ‘comp’ [-Wunused-variable]
   31 |     icalcomponent *comp = icalcomponent_get_first_real_component(ical);                                                         
      |                    ^~~~                                                                                                         
```
Consider rewriting as:
```c
assert(icalcomponent_get_first_real_component(ical))
```

rsto: I'm not following. We are using `comp` in later lines in this function and my compiler rightly so complains if I apply your suggested patch

```
libical/src/test/icalparser_ctrl_test.c:34:65: error: ‘comp’ undeclared (first use in this function)
   34 |         assert(!strcmp(want_desc, icalcomponent_get_description(comp)));
```

dilyanpalauzov: On the one side you are right, that this is not the correct fix.

On the other side there is a compiler warning, when the content of `assert()`s is stripped by the preprocessor and is invisible to the compiler.

rsto: This is test code, and other test harnesses in the same directory make use of assert, too. I think it's safe to assume that compiling it when stripping asserts is an error. If asserts should be replaced with something libical-test specific, then it would make sense to do that in general for the test code.

dilyanpalauzov: When I compile libical this is the single compiler warning I get.

dilyanpalauzov:
Undefining NDEBUG, as suggested, is a working approach.

winterz:
should be fixed with commit a5aad7152a5643c9ef891f807775f6f3a5540ebd

dilyanpalauzov:
The compiler warning is still there.  This fixes it:
```diff
diff --git a/src/test/icalparser_ctrl_test.c b/src/test/icalparser_ctrl_test.c                                                          
index c269f91b..1256b6a3 100644                                                                                                         
--- a/src/test/icalparser_ctrl_test.c                                                                                                   
+++ b/src/test/icalparser_ctrl_test.c
@@ -16,12 +16,12 @@
 #include <config.h>
 #endif
  
-#include "libical/ical.h"
-
 #if defined(NDEBUG)
 #undef NDEBUG
 #endif
  
+#include "libical/ical.h"
+
 static void assert_ctrl(enum icalparser_ctrl ctrl,
                         const char *data,
                         const char *want_desc,
```

winterz:
grr.. fixing.

winterz:
done

PR #653, opened 2023-11-09
icalparameter.c: fix RFC 6868 encoding of ^ (U+005E) character

rsto:
This fixes a bug with RFC 6868 parameter value encoding: the `^` must encoded as `^^` but was added without encoding to the parameter value instead


winterz:
could you put this in the 3.0 branch please? then I will merge into master.
also add a blurb into the ReleaseNotes.txt "Version 3.0.18" section

rsto:
Done

PR #656, opened 2023-12-21
Set LINKER_LANGUAGE for icalrecurtest

alyssais:
C++ libraries can only be reliably linked with a C++ linker. icalrecurtest is built from a C file, but links C++ libraries, so we have to tell CMake to use the C++ linker for it.

This fixes linker errors I get when attempting a static build.


winterz:
can I see the link failure?  what platform is this?

winterz:
I'm thinking about releasing a patch release for libical 3.0 soon and this would be good to add -- but I'm still not understanding where/why this is needed.   maybe platform or compiler (or compiler options ?) specific.  guess I'm asking for more info so I can try to reproduce.   has anyone else encountering this linking problem?

winterz:
I'll merge this because I don't think it can possibly do any harm.
feels like a bug in the cmake you're using though.  I mean, all the other test programs are also linking to the same library containing c++.  anyway, not worth any extra effort

PR #657, opened 2024-02-21
icalvalue.c: truncate long GEO values rather than erroring

ksmurchison:


winterz (approved):
this version works for me on Windows.  although GetNumberFormat still needs to be dealt with.  then I'm wondering why we even need to support GetNumberFormat.

ksmurchison:
Does GetNumberFormat do anything different than what we are doing manually?

ksmurchison:
> > Does GetNumberFormat do anything different than what we are doing manually?
> 
> Best I can find about GetNumberFormat is that it was introduced by me in 2011 when doing some porting for WinCE, which is long dead.
> 
> I think we can remove it in a follow-on patch and rely on what we're doing manually.

Should I remove the code that I added to that block then?

winterz:
> > > Does GetNumberFormat do anything different than what we are doing manually?
> > 
> > 
> > Best I can find about GetNumberFormat is that it was introduced by me in 2011 when doing some porting for WinCE, which is long dead.
> > I think we can remove it in a follow-on patch and rely on what we're doing manually.
> 
> Should I remove the code that I added to that block then?

yes.  change your patch to only deal with the stuff we're doing manually
then I will make a follow-on patch that removes GetNumberFormat everywhere in the code.

ksmurchison:
@winterz Done

Issue #502, opened 2021-07-30
Build system uses global CMAKE variables


paweljonskim:
Hello!
Many CMakeLists.txt use global CMAKE variables (`CMAKE_SOURCE_DIR` and `CMAKE_BUILD_DIR`) which makes impossible to add this project as subrepo in other buildsystems. Changing usage of global variables to project variables will solve the issue and should not create any additional problems.
`CMAKE_SOURCE_DIR` -> `PROJECT_SOURCE_DIR`
`CMAKE_BINARY_DIR` -> `PROJECT_BINARY_DIR`

winterz:
oops, I just created a dupe of this in #556 
I'll close that one.

PR #658, opened 2024-02-23
Fix use as a CMake sub-build

psychon:
I created a simple CMakeLists.txt in an empty directory with the following content:

    cmake_minimum_required(VERSION 3.11.0) #first line, to shutup a cygwin warning
    project(foo C) #CXX is optional for the bindings
    include(FetchContent)
    set(ICAL_GLIB FALSE)
    FetchContent_Declare(libical URL https://github.com/libical/libical/archive/refs/tags/v3.0.17.tar.gz)
    FetchContent_MakeAvailable(libical)

When trying to build this CMake project, I get the following error from "make", where SOURCE is the directory containing the above created CMakeLists.txt:

    make[2]: *** No rule to make target 'SOURCE/src/libical/icalderivedproperty.c.in', needed by 'src/libical/icalderivedproperty.c'.  Stop.

This stems from the use of the incorrect CMake variables, that always refer to the top-level source directory. By using the project-specific variables for the source directory, this can be fixed.

The actual change to the scripts was done with the following one-liner:

    git grep -l CMAKE_SOURCE_DIR  | xargs sed -e 's/CMAKE_SOURCE_DIR/PROJECT_SOURCE_DIR/g' -e 's/CMAKE_BINARY_DIR/PROJECT_BINARY_DIR/g' -i

I tested this change by patching the source in the directory _deps/libical-src that CMake created for me. A following invocation of "make" succeeded.

Fixes: https://github.com/libical/libical/issues/502


winterz:
thanks.  I meant to do this a long time ago.

PR #659, opened 2024-02-24
Buildsystem - remove support for using GetNumberFormat

winterz:
GetNumberFormat was introduced over a dozen years ago for WinCE support. This is no longer needed as we can rely on our own code for doing the same thing portably
across all platforms.


ksmurchison (approved):
LGTM

Issue #635, opened 2023-01-27
3.0.16: test suite is failing in 'icalrecurtest-r` unit


kloczek:
cmake setup (`cmake -L`)
```console
-- Cache values
CMAKE_BUILD_TYPE:STRING=RelWithDebInfo
CMAKE_INSTALL_PREFIX:PATH=/usr
ENABLE_GTK_DOC:BOOL=ON
GOBJECT_INTROSPECTION:BOOL=ON
GTKDOC_FIXXREF:FILEPATH=/usr/bin/gtkdoc-fixxref
GTKDOC_MKDB:FILEPATH=/usr/bin/gtkdoc-mkdb
GTKDOC_MKHTML:FILEPATH=/usr/bin/gtkdoc-mkhtml
GTKDOC_SCAN:FILEPATH=/usr/bin/gtkdoc-scan
GTKDOC_SCANGOBJ:FILEPATH=/usr/bin/gtkdoc-scangobj
ICAL_ALLOW_EMPTY_PROPERTIES:BOOL=ON
ICAL_BUILD_DOCS:BOOL=ON
ICAL_ERRORS_ARE_FATAL:BOOL=OFF
ICAL_GLIB:BOOL=ON
ICAL_GLIB_VAPI:BOOL=ON
ICU_DERB_EXECUTABLE:FILEPATH=ICU_DERB_EXECUTABLE-NOTFOUND
ICU_GENBRK_EXECUTABLE:FILEPATH=ICU_GENBRK_EXECUTABLE-NOTFOUND
ICU_GENCCODE_EXECUTABLE:FILEPATH=ICU_GENCCODE_EXECUTABLE-NOTFOUND
ICU_GENCFU_EXECUTABLE:FILEPATH=ICU_GENCFU_EXECUTABLE-NOTFOUND
ICU_GENCMN_EXECUTABLE:FILEPATH=ICU_GENCMN_EXECUTABLE-NOTFOUND
ICU_GENCNVAL_EXECUTABLE:FILEPATH=ICU_GENCNVAL_EXECUTABLE-NOTFOUND
ICU_GENDICT_EXECUTABLE:FILEPATH=ICU_GENDICT_EXECUTABLE-NOTFOUND
ICU_GENNORM2_EXECUTABLE:FILEPATH=ICU_GENNORM2_EXECUTABLE-NOTFOUND
ICU_GENRB_EXECUTABLE:FILEPATH=ICU_GENRB_EXECUTABLE-NOTFOUND
[...]

PR #663, opened 2024-03-16
src/test/CMakeLists.txt - create WORKING_DIR on Linux

winterz:
Needed for the icalrecurtests on linux


winterz:
fixes #635 in the master branch for linux.
I figure this is the least intrusive fix.  there are probably more elegant solutions.  
I don't know if this is also needed on Mac or Windows

Issue #604, opened 2022-10-07
Add __attribute__((sentinel)) to all _vanew function declarations


dilyanpalauzov:
The attribute ensures that the last parameter of a variadic function is `NULL`.

It was added to GCC in version 4.0, released April 2005 — https://gcc.gnu.org/gcc-4.0/changes.html

GCC definition: https://gcc.gnu.org/onlinedocs/gcc/Common-Function-Attributes.html#index-sentinel-function-attribute

The documentation of Clang 15.0 does mention the attribute at https://releases.llvm.org/15.0.0/tools/clang/docs/AttributeReference.html , the previous version does not mention it — https://releases.llvm.org/14.0.0/tools/clang/docs/AttributeReference.html .

With clang 14 the file a.c:
```c
#include <stdio.h>
static __attribute__((sentinel)) void x(int i, ...);
static __attribute__((sentinel)) void x(int i, ...)  {
  printf("Hello world %i!\n", i);
}

int main() {
  x(1, 2, 3, NULL);
  x(1, 2, 3);
  return 0;
}
```
compiled with:
```console
$ clang -Weverything a.c -o a
a.c:9:12: warning: missing sentinel in function call [-Wsentinel]
  x(1, 2, 3);
           ^
           , NULL
a.c:3:39: note: function has been explicitly marked sentinel here
static __attribute__((sentinel)) void x(int i, ...)  {
                                      ^
1 warning generated.
```
So attribute sentinel is not included in the clang 14 documentation on attributes, but is evaluated correctly by clang.

The `-Wsentinel` option is documented however as present for release clang 4.0.0, released in March 2017,  at https://releases.llvm.org/4.0.0/tools/clang/docs/DiagnosticsReference.html#wsentinel , and for previous clang versions I could not find  information on it.

winterz:
any idea about Microsoft compilers?
and then how do we deal with older clang and gcc that may not support sentinel.

could get messy. but in general I like the idea. we just need to handle the non-supported compilers

dilyanpalauzov:
I do not know, since I have no experience writing cmake recipies.

https://cmake.cmake.narkive.com/ohc1sKUc/detecting-attribute-weak-support suggests using [check_c_source_compiles()](https://cmake.org/cmake/help/v3.0/module/CheckCSourceCompiles.html) as in:
```
check_cxx_source_compiles(
"extern \"C\" void cmkcheckweak() __attribute__((weak));
int main(int argc, char** argv) {
[...]

PR #683, opened 2024-05-05
Add __attribute__((sentinel)) to all _vanew function declarations

winterz:
fixes: #604


winterz:
@dilyanpalauzov let me know what you think.

we might want to check __clang_major__ >= 14 ?  I couldn't find documentation for clang that says when sentinel attribute support was added..  

I also see no mention in the MSVC docs for a __declspec(sentinel) (or anything related)

dilyanpalauzov:
If this produces compile-time warnings, when _vanew function calls do not end with `NULL`, then the change is correct.

PR #686, opened 2024-05-09
Splint fixes

winterz:
(another patch that got lost somehow)


src/libicalvcard/vcardcomponent.c
```diff
@@ -1044,6 +1047,7 @@ static void comp_to_v3(vcardcomponent *impl)
         vcardvalue *value = vcardproperty_get_value(prop);
         vcardvalue_kind vkind = vcardvalue_isa(value);
         vcardparameter *param;
+        vcardenumarray_element type;
         char *subtype = NULL, *buf = NULL;
```
ksmurchison, line 1051: Let's just move this declaration outside the loop and we can use it throughout the function, including for the TYPE=PREF loop

ksmurchison (commented):
I don't recall seeing this patch before but LGTM with a couple of comments.
I will test on my platform now

winterz:
changes pushed according to your comments

ksmurchison (approved):

Issue #684, opened 2024-05-06
icalrecurtest-r fails with libical-3.0.18 and ICU 75.1


xry111:
test.out contains:

```
RRULE:RSCALE=RUSSIAN;FREQ=YEARLY
DTSTART:20131025
INSTANCES: *** UNIMPLEMENTED: This feature has not been implemented

RRULE:RSCALE=ETHIOPIC;FREQ=YEARLY;BYMONTH=13;BYMONTHDAY=-1;COUNT=6
DTSTART:20140910
INSTANCES: *** UNIMPLEMENTED: This feature has not been implemented

RRULE:RSCALE=CHINESE;FREQ=YEARLY;UNTIL=20180101
DTSTART:20130210
INSTANCES: *** UNIMPLEMENTED: This feature has not been implemented

RRULE:RSCALE=CHINESE;FREQ=MONTHLY;COUNT=4
DTSTART:20140920
INSTANCES: *** UNIMPLEMENTED: This feature has not been implemented

RRULE:RSCALE=ISLAMIC-CIVIL;FREQ=MONTHLY;COUNT=4
DTSTART:20131025
INSTANCES: *** UNIMPLEMENTED: This feature has not been implemented

RRULE:RSCALE=ISLAMIC-CIVIL;FREQ=YEARLY;BYMONTH=9;COUNT=5
DTSTART:20130709
INSTANCES: *** UNIMPLEMENTED: This feature has not been implemented

RRULE:RSCALE=DANGI;FREQ=YEARLY;BYMONTHDAY=8;BYMONTH=4;UNTIL=20160101
DTSTART:20131025
INSTANCES: *** UNIMPLEMENTED: This feature has not been implemented

RRULE:RSCALE=CHINESE;FREQ=YEARLY;BYMONTHDAY=10;BYMONTH=9;COUNT=3
DTSTART:20131025
INSTANCES: *** UNIMPLEMENTED: This feature has not been implemented

RRULE:RSCALE=CHINESE;FREQ=YEARLY;BYMONTHDAY=10;BYMONTH=9L;SKIP=OMIT;COUNT=2
DTSTART:20131025
INSTANCES: *** UNIMPLEMENTED: This feature has not been implemented

RRULE:RSCALE=CHINESE;FREQ=YEARLY;BYMONTHDAY=10;BYMONTH=4L;SKIP=OMIT;UNTIL=21000101
DTSTART:20131025
INSTANCES: *** UNIMPLEMENTED: This feature has not been implemented

[...]

PR #690, opened 2024-05-17
Fix build with ICU 75

lfos:
Partial fix for issue #684. Original patch by @xry111.

### DIFF
--- a/src/libical/icalrecur.c
+++ b/src/libical/icalrecur.c
@@ -1085,7 +1085,7 @@ icalarray *icalrecurrencetype_rscale_supported_calendars(void)
 
     calendars = icalarray_new(sizeof(const char **), 20);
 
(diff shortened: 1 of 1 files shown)


winterz:
this patch is against the master branch.
we want this in the 3.0 branch.  then I 'll merge it to the master branch.

you say "partial fix".  what's missing?

lfos:
> this patch is against the master branch. we want this in the 3.0 branch. then I 'll merge it to the master branch.

My apologies, I am used to development workflows where PRs are made against the master branch, and then backported to stable branches from there. Happy to resend as a patch for the 3.0 branch if preferred. In that case, should it have the conditional you proposed in issue #684?

> you say "partial fix". what's missing?

This fix is self-contained; "partial" because there's another unrelated issue with ICU 75 (see #684), so it doesn't fully fix compatibility with ICU 75.

PR #693, opened 2024-05-28
icalcomponent.c: do not dereference NULL when comparing components

rsto:
The icalcomponent_normalize function sorts components by comparing their mandatory property values. This causes it to dereference a NULL pointer if one of the compared components erroneously is missing such a property.

This patch fixes that by checking for the presence of mandatory properties before comparison, otherwise sorting components having a property before ones that don't.


rsto:
Now with fixed clang formatting.

PR #700, opened 2024-06-05
WIP - fix some more fuzzer issues

winterz:
icalmemory.c - icalmemory_tmp_copy(): don't allocate memory
for an empty string (i.e a string that starts with '\0')

icalparser.c - icalparser_add_line(): END is encountered
before the BEGIN becomes an error, not just a warning

icaltypes.c -  icalreqstattype_from_string(): don't allow
garbage (non-printable chars) in the reqstattype debug field.
for unknown reasons the icalmemory tmp memory ring doesn't
work properly on garbage; additionally the snprintf in
icalreqstattype_as_string_r() uses the "%s" format which
is not intended for unprintable chars.


src/libical/icaltypes.c
```diff
@@ -130,7 +130,18 @@ struct icalreqstattype icalreqstattype_from_string(const char *str)
 
     p2 = strchr(p1 + 1, ';');
     if (p2 != 0 && *p2 != 0) {
+#if 0
         stat.debug = icalmemory_tmp_copy(p2 + 1);
+#else
+        // Don't allow (fuzzer) garbage else somethine in icalmemory fails
```
ksuther, line 136: Typo: somethine -> something

src/libical/icaltypes.c
```diff
@@ -130,7 +130,18 @@ struct icalreqstattype icalreqstattype_from_string(const char *str)
...
+        char *s = (char *)(p2 + 1);
+        while (*s && isprint((unsigned char)*s)) {
```
ksuther, line 138: What about moving this up to the top of the function and failing immediately if it encounters garbage in `str`? We're reading from `str` higher up, so seems like we should be checking earlier too.

ksuther (commented):
fuzz1test69344 and fuzz2test69344 are segfaulting now, is that expected?

The changes seems reasonable to me, other than the isprint check might be too late. I'm assuming the `#if 0` portions will go away for the final version?

ksmurchison:
Im on the last day of a cruise and then spending a couple of more days in Europe. I probably won't get to this until Monday.

winterz:
> fuzz1test69344 and fuzz2test69344 are segfaulting now, is that expected?
> 
no.
afaict those are happening on Windows only and so far I'm stumped as to why.  I'm not a debugging wizard on Windows.

> The changes seems reasonable to me, other than the isprint check might be too late.
yes, I'll look at moving the "check for unprintable chars" to higher up in the function.

> I'm assuming the `#if 0` portions will go away for the final version?
yes.

winterz:
this is ready for another review.
I removed the fuzzing tests on windows for now. 1 of the fuzzer tests failed on windows and I couldn't figure out why.

ksuther (approved):

Issue #637, opened 2023-03-24
Ignoring RDATE;VALUE=PERIOD parameter


dubltwoo:
https://www.rfc-editor.org/rfc/rfc5545#section-3.8.5.2

>BEGIN:VEVENT
DTSTART:20230324T080000Z
DURATION:PT1H
RRULE:FREQ=DAILY;UNTIL=20230328T230000Z
RDATE;VALUE=PERIOD:20230325T120000Z/PT3H
END:VEVENT

According to the documentation, i should get time span like this

>start: 2023-03-24T08:00:00Z end: 2023-03-24T09:00:00Z
start: 2023-03-25T08:00:00Z end: 2023-03-25T09:00:00Z
**start: 2023-03-25T12:00:00Z end: 2023-03-25T15:00:00Z**
start: 2023-03-26T08:00:00Z end: 2023-03-26T09:00:00Z
start: 2023-03-27T08:00:00Z end: 2023-03-27T09:00:00Z
start: 2023-03-28T08:00:00Z end: 2023-03-28T09:00:00Z

but i get without 
>**start: 2023-03-25T12:00:00Z end: 2023-03-25T15:00:00Z**

minichma:
You have a typo in `PERIOND`. Does it work if you fix it?

dubltwoo:
Sorry for typo. I have corrected
> Does it work if you fix it?
No, it dose not work

ksmurchison:
This is a known limitation.  Comment from icalcomponent.c:icalcomponent_foreach_recurrence()

```
        /* RDATES can specify raw datetimes, periods, or dates.
            we only support raw datetimes for now..

            @todo Add support for other types */

```

CMendia:
@winterz I think this one was actually already fixed. I added a test for date-value RDATEs in #981 .

PR #707, opened 2024-06-08
icalcomponent_foreach_recurrence(): add support for RDATE;VALUE=PERIOD

ksmurchison:
Potential fix for issue #637.  I did this on an airplane so it needs to be sanity checked.

### DIFF
--- a/src/libical/icalcomponent.c
+++ b/src/libical/icalcomponent.c
@@ -957,25 +957,40 @@ void icalcomponent_foreach_recurrence(icalcomponent *comp,
          rdate = icalcomponent_get_next_property(comp, ICAL_RDATE_PROPERTY)) {
         struct icaldatetimeperiodtype rdate_period =
             icalproperty_get_rdate(rdate);
+        struct icaltimetype rdate_start = rdate_period.time;
+        icaltime_t rdate_duration = 0;
 
         /* RDATES can specify raw datetimes, periods, or dates.
-            we only support raw datetimes for now..
+            we only support raw datetimes and periods for now.
 
             @todo Add support for other types */
 
-        if (icaltime_is_null_time(rdate_period.time))
-            continue;
+        if (icaltime_is_null_time(rdate_start)) {
+            rdate_start = rdate_period.period.start;
+
+            if (icaltime_is_null_time(rdate_period.period.end)) {
+                rdate_duration =
+                    (icaltime_t)icaldurationtype_as_int(rdate_period.period.duration);
+            } else {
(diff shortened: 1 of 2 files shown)


winterz:
Splint issue:

```
src/libical/icalcomponent.c: (in function icalcomponent_foreach_recurrence)
src/libical/icalcomponent.c:972:17: Assignment of int to time_t: rdate_duration = icaldurationtype_as_int(rdate_period.period.duration)
  To allow arbitrary integral types to match long unsigned, use +longintegral.
```

winterz (commented):
looks good except for the 1 splint issue.
wouldn't hurt to add some more test cases to regression.c.

ksmurchison:
> looks good except for the 1 splint issue. wouldn't hurt to add some more test cases to regression.c.

@winterz What other tests would you suggest?

winterz:
> > looks good except for the 1 splint issue.
> 
> Does typecasting the value to icaltime_t solve the issue. E.g.: rdate_duration = (icaltime_t) icaldurationtype_as_int(...)
> 
yep

> > wouldn't hurt to add some more test cases to regression.c.
> 
> @winterz What other tests would you suggest?
* starttime with duration (rather than endtime)
*  the endtime is before the starttime (badness)


btw: should
 "RDATE;VALUE=PERIOD:20180223T020000Z/20180223T030000Z," 
be
 "RDATE;VALUE=PERIOD:20180223T020000Z/20180223T030000Z\n"

ksmurchison:
> > > looks good except for the 1 splint issue.
> > 
> > 
> > Does typecasting the value to icaltime_t solve the issue. E.g.: rdate_duration = (icaltime_t) icaldurationtype_as_int(...)
> 
> yep
> 
> > > wouldn't hurt to add some more test cases to regression.c.
> > 
> > 
> > @winterz What other tests would you suggest?
> 
>     * starttime with duration (rather than endtime)
> 
>     * the endtime is before the starttime (badness)

Neither of these really impact RDATE;VALUE-PERIOD handling specifically.  They would be more appropriate for general recurrence handling.

> btw: should "RDATE;VALUE=PERIOD:20180223T020000Z/20180223T030000Z," be "RDATE;VALUE=PERIOD:20180223T020000Z/20180223T030000Z\n"

Only if we adding folding whitespace following the linefeed.  I'm fine doing that if you prefer it.

winterz:
ok.  then good to go after fixing that icaltime_t cast

ksmurchison:
Done

PR #717, opened 2024-06-25
icalvalue.c - don't allow huge, garbage data to cause a hang

winterz:
In icalmemory_strdup_and_quote() don't keep looping over
garbage data for a very long time. 10000 times should be plenty.

Fix for fuzzer 69888


winterz:
I think this is a safe hack. 
there is no other way I can see to detect if there is a large amount of garbage data

I'm interested in other opinions.

ksuther (approved):
My only nitpick would be that I prefer doing `static const size_t MAX_ITERATIONS = 10000;` instead of a `#define` to get type checking. I know we do plenty of `#define` elsewhere already though, so feel free to ignore me :)

Last time I was dealing with fuzzer failures I set a max number of properties and parameters, which is essentially the same approach.

Can an entire `DESCRIPTION` get passed to this function or is it chunked? I've seen multi-megabyte emails packed into `DESCRIPTION` in the past, which would suggest setting the max size to be even larger (something like `1024 * 1024 * 10`).

winterz:
> My only nitpick would be that I prefer doing `static const size_t MAX_ITERATIONS = 10000;` instead of a `#define` to get type checking. I know we do plenty of `#define` elsewhere already though, so feel free to ignore me :)
> 
> Last time I was dealing with fuzzer failures I set a max number of properties and parameters, which is essentially the same approach.
> 
> Can an entire `DESCRIPTION` get passed to this function or is it chunked? I've seen multi-megabyte emails packed into `DESCRIPTION` in the past, which would suggest setting the max size to be even larger (something like `1024 * 1024 * 10`).

agreed.  force pushed those changes

ksmurchison (approved):

PR #725, opened 2024-09-14
Compile the project with -Werror (non-Windows)

winterz:


winterz:
Anyone with a Mac that could play with this patch and tell us how to avoid

```
libical-glib-scan.c:601:30: error: unused parameter 'fp' [-Werror,-Wunused-parameter]
output_object_actions (FILE *fp, GType object_type)
                             ^
```
this patch works on linux with gcc and clang.  I don't have access to a Mac

PR #733, opened 2024-09-18
.github/workflows/build.yml - move the Mac runner to macos-14

winterz:


winterz:
giving up
if any Mac people are watching => help

PR #742, opened 2024-09-23
icaltime.c - icaltime_is_valid_time() also validate month and day

winterz:
If the month and day are out-of-range then the time is invalid

### DIFF
--- a/src/libical/icaltime.c
+++ b/src/libical/icaltime.c
@@ -587,7 +587,10 @@ struct icaltimetype icaltime_null_date(void)
 
 int icaltime_is_valid_time(const struct icaltimetype t)
 {
-    if (t.year < 0 || t.year > 3000 || t.is_date > 1 || t.is_date < 0) {
+    if (t.year < 0 || t.year > 3000 ||
+        t.month < 0 || t.month > 12 ||
+        t.day < 0 || t.day > 31 ||
+        t.is_date > 1 || t.is_date < 0) {
         return 0;
     } else {
         return 1;


ksmurchison (commented):
Looks sane, but a couple of comments?

- Is there a tech reason why we don't allow year > 3000?  RFC 5545 allows up to 9999 and some clients will set the ACKNOWLEDGED property value having a year of 9999.
- Should we also be sanity checking the time parts?

rsto:
Some more drive-by comments:

- Should this also validate if a day of month actually exists for that month, including leap years?
- Last time I used `icaltime_is_valid_time` I found that function name confusing: does it only check for the time parts? Only after I checked the code I figured out what it actually does (and does not). Should we refactor this to `icaltime_is_valid_date` (which checks the date part), `icaltime_is_valid_time`(which checks the time part) and `icaltime_is_valid_datetime` which combines both?

I can contribute code for this, if there is interest.

winterz:
> Some more drive-by comments:
> 
>     * Should this also validate if a day of month actually exists for that month, including leap years?
> 
>     * Last time I used `icaltime_is_valid_time` I found that function name confusing: does it only check for the time parts? Only after I checked the code I figured out what it actually does (and does not). Should we refactor this to `icaltime_is_valid_date` (which checks the date part), `icaltime_is_valid_time`(which checks the time part) and `icaltime_is_valid_datetime` which combines both?

the purpose of `icaltime_is_valid_time` should be "is the data stored in the icaltimetype structure valid"?
unfortunately, that also implies renaming `struct icaltimetype` into `struct icaldatetimetype` but that would be a huge change.  

iow:
 - rename `struct icaltimetype` to  `struct icaldatetimetype`
 - rename `icaltime_is_valid_time` to `icaltime_is_valid_datetime`
- find all API where we use "time" that should be renamed to "datetime" 

But who wants to do all that work. Not me.
In the meantime, I think your proposal is a good idea and would like to have a patch

PR #746, opened 2024-09-24
libical-glib - Make sure to always free a icalmemory buffer with the appropriate function

### LINKED ISSUE #744: Investigate why the Python bindings have memory failures on Mac with the MEMORY_CONSISTENCY check
see https://github.com/libical/libical/issues/738
> The `MEMORY_CONSISTENCY` is also something interesting to me. Maybe python does some magic with the strings and copying them from one (python) object to another can garbage the logic behind this option. I do not know python, this is only a pure guess from my side. I do not know a gobject-introspection tag to note that "this memory can be copied and freed only with this and this function". The non-basic types (like structures) can be registered a `GBoxed` types, which do have these copy & free functions defined, but the string (`char *`) is a basic type and trying to register it as a boxed type might break other things.

I'll create a new issue -- maybe we can find a Mac and/or Python expert to help out.  I don't know much about Python either.  Could be that we need to use the official Python for Mac rather than the python provided by homebrew.  I really don't know.

tintou:
Closes #744

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -670,12 +670,6 @@ endif()
 
 libical_option(LIBICAL_DEVMODE_MEMORY_CONSISTENCY "(Developer-only) Build with memory consistency functions." False)
 if(LIBICAL_DEVMODE_MEMORY_CONSISTENCY)
-  if(GOBJECT_INTROSPECTION)
-    message(
-      FATAL_ERROR "Memory consistency combined with GObject Introspection is incompatible at this time.\n"
-                  "Please disable either the LIBICAL_DEVMODE_MEMORY_CONSISTENCY or GOBJECT_INTROSPECTION CMake options."
-    )
-  endif()
   set(CMAKE_BUILD_TYPE "Debug")
   add_definitions(-DMEMORY_CONSISTENCY)
 endif()
--- a/scripts/mkderivedparameters.pl
+++ b/scripts/mkderivedparameters.pl
@@ -308,7 +308,7 @@ sub insert_code
             $charorenum =
                 "    icalerror_check_arg_rz((param != 0), \"param\");\n    return param->string;";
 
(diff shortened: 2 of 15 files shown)


mcrha:
I agree there are not many options here, but the "some overheads" from the second commit message is kinda inaccurate from my point of view. This can cause a giant memory fragmentation, not talking about the performance hit.

Would it make sense to duplicate the string only if the `MEMORY_CONSISTENCY` developer option is in use? Users/distros are not meant to use this developer-only compiler option, thus penalize them is not good. I know the conditional compile will "garbage" the libical-glib tests, because what the users have and what the special build runs would be different, but I'm afraid the penalty is too high.

Just my opinion.

tintou:
Another option would be for all the functions to return a `ICalString *` that would be a typedef to `gchar *` and defined as boxed type, but this is technically an API break

mcrha:
> but this is technically an Introspection API break

not only introspection, but also C API break, a giant one (if nothing else, it could produce an incompatible pointer type warning from the compiler, which many distros turn into an error and abort the build).

Again, the `MEMORY_CONSISTENCY` option is only for the developers, not for the users/distros. One can just call libical-glib "untestable" in such environment and consider it a known limitation. I'd not call it a known bug.

minichma:
> Would it make sense to duplicate the string only if the MEMORY_CONSISTENCY developer option is in use?

I think, this would be a viable option and the best compromise. And it could easily be applied to this PR, which looks good to me aside from this topic.

mcrha:
I gave this branch some smoke testing and I do not see any significant penalty. I thought there will be some, but really nothing noticeable, at least here. Also the memory fragmentation I've been afraid of did not strike anything significant in compare with the master branch without these changes. Might be I'm simply wrong or it's just the difference between the theory and the practice (most likely the former). I'm sorry for the noise.

winterz:
I'm giving up on the Mac building with GOBJECT_INSPECTION.  There must be more homebrew packages needed.  will try again in another pull request

Issue #387, opened 2019-05-13
API passbyvalue


winterz:
Coverity is (rightly) complaining about some functions that take moderately sized structures where we it would be better to pass pointers to those structures.

in particular we pass struct icalrecurrencetype in quite a few places.
this would be an API change, unfortunately.   in 3.1 we should deprecate the bad signatures and remove those in 4.0

look for all functions where we pass structs and not pointers to structs and try to be consistent about it.

minichma:
Copied over from duplicate #548:

The `icalrecurrencetype` struct is passed throughout libical by value. This probably has historic reasons, as it doesn't require much memory management and used to be sufficient. But with introduction of rscale, the new `icalrecurrencetype.rscale` attribte was introduced, which points at a dynamically allocated block of memory. The benefit of easy memory management is gone since then, as anyone using the type now needs to manage the referenced memory.

So this is to discuss, whether it would make sense to treat `icalrecurrencetype` as a reference type. This would mean introducing funcs like `icalrecur_new()`, `icalrecur_free()` and `icalrecur_clone()`. Cloning and freeing the rscale memory would be done inside `icalrecur_clone()` and `icalrecur_free()`, which would improve encapsulation and decoupling.

Another benefits would be, that the type could be made more dynamic as a whole over time. Today the size of the type is quite large, which can (and does) cause issues on resource-constrained devices, like embedded ones. By making it a reference type and centralizing the type's memory management within `icalrecur.c`, we could easily make it more dynamic. E.g. we could replace the big static by_xxx arrays with dynamic ones, therefore saving significant amounts of memory (in most cases).

The major downside would be backwards-compatibility. I.e. `icalproperty_new_rrule()` et al. would change their signature, as they would have to take pointers instead of values. Are these functions considered public? This would probably require introducing this only in a new major version, right?

We implemented something like this in a fork foccussed at embedded devices. The main changes (based on a version from 2018) can be seen here (including both, the change to a reference type as well as replacing the static by_xxx arrays):
https://github.com/tapkey/libical/compare/5683dd4..9579d55

If considered beneficial, I'd consider refurbishing the change and contributing it back to libical.

minichma:
If we want to have both versions (byref and byval) at least between versions 3.1 and 4.0, we need to find new names for the new versions. Which names would be good ones?

`icalrecur_iterator_new ` -> `icalrecur_iterator_new_r`, `icalrecur_iterator_new_byref`, ...?
[...]

Issue #202, opened 2015-06-14
Perl Bindings


winterz:
Bring back the Perl bindings (see the src/Net-ICal-Libical folder)

Issue #199, opened 2015-06-14
Java Bindings


winterz:
bring back the Java bindings (in the src/java folder)

PR #751, opened 2024-09-26
Make icalrecurrencetype a type that is passed by ref rather than by val.

minichma:
Changes `icalrecurrencetype` to be passed by reference rather than by value. The concept is very close to how `icalattach` works. I.e.
* Instances of `struct icalrecurrencetype` are created `icalrecurrencetype_new()` or `icalrecurrencetype_new_from_string()`, which allocates them on the heap, rather than on the stack. New objects are returned with a refcount of 1.
* Memory is managed via reference counting (via `icalrecurrencetype_ref()` and `icalrecurrencetype_unref()`). Objects are automatically freed when the reference counter reaches 0. There is no explicit `icalrecurrencetype_free()` function.

The GLib API stays unchanged, except for the `i_cal_recurrence_clear()` method, which has been removed, as it doesn't seem to make too much sense now. The C version `icalrecurrencetype_clear` was mainly used to mark an object as undefined, which is now done by setting the pointer to NULL, so it has been removed. But that's TBD.

ToDo
- [x] Review GLib binding. I don't know too much about GLib in general and our bindings in particular, this should be reviewed carefully. Especially the initial reference counting and relation to the object's owner (compare the binding of icalattach, where we have this `new_full_extra_code="if (owner) { icalattach_ref (native); owner = NULL; }"`.
- [x] Update documentation and code comments.
- [x] Release notes.

Fixes https://github.com/libical/libical/issues/387


minichma:
@winterz What do you think? Who would be the one having a closer look at the GLib binding? I'm not an expert there, so it should be reviewed carefully.

winterz (commented):
looking now. 
I fixed up the .md file a bit and am currently working in a new branch called minichma-recur_byref

minichma:
Great, thank you. Added two minor fixes onto minichma-recur_byref that I just came across, not too relevant. Should be squashed before merging.

winterz:
we don't care about Java bindings at this time. (see issue #199)
nor Perl (#202)

you can remove the Java bindings from this issue's task list

winterz:
so far I can't find any problems with glib implementation.
I think we go ahead with it and deal with any unforeseen problems in follow-on patches

minichma:
Alright, should I reset this PR onto the minichma branch? And start a 4.0 section in the release notes?

winterz:
> Alright, should I reset this PR onto the minichma branch?

yes.

>  And start a 4.0 section in the release notes?
hmm.. right, we need to rename "3.1" to "4.0" in the ReleaseNotes
for now just add to the list under "Version 3.1.0 (NOT RELEASED YET):" .  I'll eventually change that to "Version 4.0.0"

PR #766, opened 2024-10-03
icalrecurrencetype: Replace static 'by' arrays with dynamically allocated ones.

minichma:
Now that `icalrecurrencetype` is [passed by ref rather than by value](https://github.com/libical/libical/pull/751), we can continue by making the type more dynamic. This PR replaces the 'by_xxx' arrays that have previously been allocated as statically sized arrays as part of `icalrecurrencetype`, with dynamically allocated ones. This allows to significantly reduce the type's memory footprint (in all real use cases), which used to be > 2k. As a consequence of the dynamic arrays, we can also improve the code in many places by making it more generic.

The PR does:
- Replace statically allocated 'by_xxx' arrays with dynamically allocated ones.
- Replace 'by_xxx' fields with a single, more generic `by[ICAL_BY_XXX]` array
- Simplify the GLib/Python bindings by replacing the individual '*_by_xxx' methods with generic ones that take the 'by' component as an enum parameter.
- Avoid cloning the rule in `icalrecurrencetype_iterator_new()` if not necessary (i.e. in all cases except if rscale is used).

Todos
- [x] Discuss modified GLib API (i.e. replace individual 'by_xxx' methods with more generic, parameterized ones)
- [x] Documentation & Release notes

### DIFF
--- a/ReleaseNotes.txt
+++ b/ReleaseNotes.txt
@@ -21,6 +21,9 @@ Version 3.1.0 (NOT RELEASED YET):
  * libical-glib API is considered stable; no longer need to define LIBICAL_GLIB_UNSTABLE_API=1
    before including <libical-glib/libical-glib.h>
  * icalrecurrencetype now handled by reference rather than by value throughout the library.
+ * icalrecur_iterator_new() now keeps a reference to the passed rule, so that mustn't be modified as long as the iterator is in use.
+ * icalrecurrencetype.by_xxx static arrays replaced by dynamically allocated ones.
+ * GLib/Python bindings: `i_cal_recurrence_*_by_xxx*` methods have been replaced by more generic versions that take the 'by' type (day, month, ...) as a parameter.
  * Allow previous recurrence iteration
  * Improved performance of recurrence iterators
  * GEO property has arbitrary precision (values are internally stored as strings, not doubles)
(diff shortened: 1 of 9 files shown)


winterz:
cppcheck reports:
```/home/allen/projects/libical/libical/src/libical/icalrecur.c:2394,error,negativeIndex,Array 'impl->bydata[9]' accessed at index -1, which is out of bounds.```

winterz:
looking really good...

minichma:
Thank you, glad you like it! So I will continue working on it. I.e. have another close look and add missing docs/release notes. So you think, we can change the GLib/Python API as proposed? I believe it should be nicer to use also for the consumers.

winterz:
> So you think, we can change the GLib/Python API as proposed? I believe it should be nicer to use also for the consumers.

I agree.

minichma:
Should be ready for review, except for the markdown linter complaints, which I'll try to fix tomorrow.

winterz (approved):

ksmurchison:
Not a comment on this PR but about icalrecur-by-ref in general.  Should icalproperty_get_rrule/exrule() be bumping the ref count?  If I do a icalproperty_get_rrule() followed by a icalproperty_set_rrule() using what I just retrieved, the icalrecurrencetype will be freed from icalvalue before it gets set.

ksmurchison:
> @ksmurchison Is this still an issue? What do you think about the way refcounting is implemented?

Not an issue.  I just completed writing my compatibility layer to support both icalrecur-by-ref and icalrecur-by-value versions of libical and the current ref counting works as expected.  My code bumps the refcount after icalproperty_get_rrule() and the decrements it when doing using it

minichma:
@winterz How should we proceed from here? Can we merge?

winterz:
sure. merge at will

PR #784, opened 2024-10-10
icalrecur.c: icalrecurrencetype_clone() should initialize its refcount

ksmurchison:
Since this is a fresh copy of an existing icalrecurrencetype, we SHOULD NOT inherit the refcount.  This prevents us from leaking the icalrecurrentype inside an icaliterator


minichma:
Oh yes, for sure. Thanks for fixing!

minichma (approved):

PR #793, opened 2024-10-22
Test: Parameterize icalrecur_test

minichma:
Remove test case definitions from `icalrecur_test.c` in favor of parsing the related `icalrecur_test*.out` files and verify success within `icalrecur_test.c` rather than via an external file comparison.

`icalrecur_test` used to work by generating an output file based on test cases hard-coded in the source code. The output was then compared in a separate step, by a separate utility for equality with reference data. By moving the test definitions out of icalrecur_test and moving the comparison for success in, we gain following improvements:

* Tests are easier to write and maintain, as test cases are only maintained in a single place (.out file) rather than in two (source file and .out file).
* Tests are easier to debug, as we know, whether test succeed after only running icalrecur_test, without having to do an additional file comparison. We can also work with breakpoints more easily.


ksmurchison:
I like this approach a lot, but I'd like to find a way to keep the comments for each test as I feel they add an important context to some of them.

minichma:
@winterz Would you be available for a review? What do you think of this approach?

winterz:
I'm back from a few days vacation and will try to catch up this weekend

winterz:
/home/allen/projects/libical/libical/src/test/icalrecur_test.c:168:17: warning: Value stored to 'actual_instances_len' is never read [deadcode.DeadStores]
/home/allen/projects/libical/libical/src/test/icalrecur_test.c:236:13: warning: Value stored to 'yield' is never read [deadcode.DeadStores]

minichma:
@winterz Thanks for the review! Fixed the warnings. I assume the failing macos build is not related to this PR, right?

winterz:
> @winterz Thanks for the review! Fixed the warnings. I assume the failing macos build is not related to this PR, right?

right.
we're good to go

winterz (approved):

winterz (approved):

PR #796, opened 2024-10-24
Recurrence: Allow limiting BY rules using negative values.

### LINKED ISSUE #791: `FREQ=HOURLY;BYYEARDAY=-366` doesn't return any recurrences
Added the following test case to `icalrecur_test.c`:

```c
    /* Test hourly with -largest day of year as limit */
    {"20120101T120000",
     "FREQ=HOURLY;BYHOUR=12;BYYEARDAY=-366;COUNT=3",
     NULL},
```

The case doesn't generate any recurrence instances. Expected the same as in the test case

```c
    /* Test yearly with -largest day of year */
    {"20120101T120000",
     "FREQ=YEARLY;BYYEARDAY=-366;COUNT=3",
     NULL},
```

I.e. `INSTANCES:20120101T120000,20160101T120000,20200101T120000`

minichma:
Allow limiting using negative BY rules. I.e. support the following patterns:

`FREQ=[DAILY|HOURLY|MINUTELY|SECONDLY];BYMONTHDAY=-X`
`FREQ=[HOURLY|MINUTELY|SECONDLY];BYYEARDAY=-X`

Fixes #791

This is blocked by #792 and #793 and needs to be rebased after they are merged. Only the last two commits belong to this PR.


ksmurchison:
This is great work.  

However, I was thinking earlier today that the entire recurrence rule expansion could use (another) complete rewrite, potentially using the algorithm outlined in [https://www.rfc-editor.org/rfc/rfc8984.html#section-4.3.3.1].  One of my work colleagues already has this implemented in Javascript for our Fastmail UI and in Rust as a pet project.  I think this would greatly simplify the code and possibly cover all of the BY* combinations that we currently don't support.  The only tricky part would be implementing it both with and without ICU.  But if we factor out the right pieces like we currently have, it should be straightforward.

Having said that, I'm going to be away from my desk for the next two weeks on a weekend holiday in Scotland, followed by a work conference in London and then another in Dublin.  If I have time I might try to tinker a bit.

minichma:
I appreciate the JSCalendar efforts a lot and think aligning with it would generally be a great idea. I had the pleasure to meet Robert, who I believe is a colleague of yours, and chat with him here in Vienna. Should you happen to visit Vienna on your Europe trip, we could have a beer together :-).

I don't know enough about JSCalendar to know, whether they are fully compatible though. Anyhow, applying this fix in the meantime would probably be a good idea nevertheless. Note that only the last two commits are specific to this issue. The rest comes from the other PRs that are currently open. Anyhow, I'm pretty sure this can easily wait for two weeks.

ksmurchison:
Yes. Robert is my colleague. 
I plan to be in Vienna in July 2026 for IETF. Let's plan on a beer or 3

winterz (approved):
I'm happy to have this unless/until there's a rewrite.
one small thing: please add a blurb to the ReleaseNotes.txt that "limiting BY rules using negative values" is now supported

winterz (approved):

minichma:
Thanks for reviewing! Now that we have the `KnownLimitations.md` I think we don't need to back-port this to 3.0. What do you think?

winterz:
yes, cherry-pick to 3.0 would be great

Issue #798, opened 2024-10-26
Document Known Limitations


winterz:
refer to #795 

document libical's known limitations in docs/KnownLimitations.md
like to this in docs/SUMMARY.md so the known limitations appear in the Libical book

PR #799, opened 2024-10-27
docs/KnownLimitations.md - start to document known limitations

winterz:
New to document missing features in Libical.

fixes: #798

### DIFF
--- a/docs/KnownLimitations.md
+++ b/docs/KnownLimitations.md
@@ -1,0 +1,22 @@
+# Known Limitations
+
+List where libical doesn't fully support all the features described in the RFCs.
+
+Contributions to help us fix these limitations are welcome.
+
+## RFC 5545
+
+### Recurrence Rule section 3.3.10
+
+* BYSETPOS is only supported for MONTHLY and YEARLY frequencies (<https://github.com/libical/libical/issues/795>)
+
+* When used as a limit (rather than for expansion) negative values are not supported
+  for BYMONTHDAY, BYYEARDAY (<https://github.com/libical/libical/issues/791>)
+
+* FREQ=YEARLY, BYYEARDAY can't be combined with BYWEEKNO, BYMONTH or BYMONTHDAY
+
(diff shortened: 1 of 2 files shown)


docs/KnownLimitations.md
```diff
@@ -0,0 +1,47 @@
...
+Contributions to help us fix these limitations are welcome.
+
+## RFC 5545
+
+### Recurrence Rule section 3.3.10
+
+* BYSETPOS is only supported for MONTHLY and YEARLY frequencies (<https://github.com/libical/libical/issues/795>)
+
+* When used as a limit (rather than for expansion) negative values are not supported
+  for BYMONTHDAY, BYYEARDAY. possibly others (<https://github.com/libical/libical/issues/791>)
```
minichma, line 14: Probably 'possibly others' is not too helpful for end users and I don't think there are others. Comment was there more for internal discussion. Would remove it.

docs/KnownLimitations.md
```diff
@@ -0,0 +1,47 @@
...
+
+* FREQ=YEARLY, BYYEARDAY can't be combined with BYWEEKNO, BYMONTH or BYMONTHDAY
+
+     ref: <https://github.com/libical/libical/blob/cfd401b9d043214395888de1d9daf52263e3245b/src/libical/icalrecur.c#L2882>
+
+     ```c
```
minichma, line 20: Don't think we need to add the code snippet here. By adding the link, users are free to look it up if they're interested.

docs/KnownLimitations.md
```diff
@@ -0,0 +1,47 @@
...
+              return 0;
+          }
+     )
+     ```
+
+* FREQ=YEARLY, BYWEEKNO can't be combined with BYYEARDAY, BYMONTH or BYMONTHDAY
+
+     ref: <https://github.com/libical/libical/blob/cfd401b9d043214395888de1d9daf52263e3245b/src/libical/icalrecur.c#L2928>
+
+     ```c
```
minichma, line 35: Same as above - wouldn't add the snippet.

minichma (commented):
Looks good, I think this will be helpful (would have helped us).

Issue #723, opened 2024-08-19
Error on macOS 14: `libical-glib-private.h:119:2: error: #endif without #if`


barracuda156:
Got a weird error on macOS 14 CI:

```
  In file included from /opt/local/var/macports/build/_Users_runner_work_macports-ports_macports-ports_ports_devel_libical/libical/work/build/src/libical-glib/i-cal-attach.c:24:
  /opt/local/var/macports/build/_Users_runner_work_macports-ports_macports-ports_ports_devel_libical/libical/work/build/src/libical-glib/libical-glib-private.h:119:2: error: #endif without #if
  #endif
   ^
  1 error generated.
  make[2]: *** [src/libical-glib/CMakeFiles/ical-glib.dir/i-cal-attach.c.o] Error 1
  make[2]: Leaving directory `/opt/local/var/macports/build/_Users_runner_work_macports-ports_macports-ports_ports_devel_libical/libical/work/build'
  make[1]: *** [src/libical-glib/CMakeFiles/ical-glib.dir/all] Error 2
```

Log: https://github.com/macports/macports-ports/actions/runs/10446373347/job/28923692179?pr=25391

winterz:
I don't know what could be wrong but it does bring up the point that we need to support more platforms in our .github/workflows /build.yml

if anyone knows how, would be nice to add macos-13 and macsos-14.  and any other useful platforms

barracuda156:
MacPorts has CI on those, perhaps the setup can be just copied? And many repos do. Current MacOS on mainstream archs aren’t really exotic.

mcrha:
I sometimes get this error, a simple rebuild usually helps (usually one, but...).

I believe the dependencies of the custom commands are set properly, but it seems they are not. I think the problem is that the build runs several parts of the library in parallel and for some reason also the libical-glib's API files transform from .xml to .c/.h files sometimes happens at the same time, effectively overwriting the `build/src/libical-glib/libical-glib-private.h` in the middle of the write from another thread/process/whatever CMake calls it when it runs the build in parallel. it's a timing issue, because the other time it is fine.

I do not know how to fix this properly. Thinking of it now, maybe the custom command to create those files may create some file not directly, but through a temporary file and rename that temporary file only after the content of it is fully populated. But whether it's it, I do not know.

You should be able to workaround the problem by running `make -j1` or `ninja -j1` or a similar option in the build command you configure the project with. Make sure you clean up the build directory before it, at least the generated libical-glib files.

[...]

PR #805, opened 2024-11-06
Fix several libical-glib issues

mcrha:
Please see the respective commits.

### DIFF
--- a/docs/reference/libical-glib/libical-glib-docs.sgml.in
+++ b/docs/reference/libical-glib/libical-glib-docs.sgml.in
@@ -60,23 +60,23 @@
     <title>Index of new symbols in 4.0</title>
     <xi:include href="xml/api-index-4.0.xml"><xi:fallback /></xi:include>
   </index>
-  <index id="api-index-3-0-5" role="3.0.5">
+  <index id="api-index-3-0-5">
     <title>Index of new symbols in 3.0.5</title>
     <xi:include href="xml/api-index-3.0.5.xml"><xi:fallback /></xi:include>
   </index>
-  <index id="api-index-3-0" role="3.0">
+  <index id="api-index-3-0">
     <title>Index of new symbols in 3.0</title>
     <xi:include href="xml/api-index-3.0.xml"><xi:fallback /></xi:include>
   </index>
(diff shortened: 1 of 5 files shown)


tintou (commented):
vapigen does not use the typelib, how did you trigger this issue?

tintou:
other commits are RB

mcrha:
> vapigen does not use the typelib, how did you trigger this issue?

Hmm, does it not? Weird, I though it does.

I do not know what causes it, some maybe system setting, or such, I only faced it while playing with these changes. When I disable parallel build it works. When I add the typelib dependency it works. Without it I _sometimes_ get during `make && make install`:
```
libical-glib-scan.c: In function ‘output_object_actions’:
libical-glib-scan.c:601:30: warning: unused parameter ‘fp’ [-Wunused-parameter]
  601 | output_object_actions (FILE *fp, GType object_type)
      |                        ~~~~~~^~

make[2]: *** [src/libical-glib/CMakeFiles/valafile.dir/build.make:75: src/libical-glib/libical-glib.vapi] Bus error (core dumped)
make[1]: *** [CMakeFiles/Makefile2:1080: src/libical-glib/CMakeFiles/valafile.dir/all] Error 2
make[1]: *** Waiting for unfinished jobs....
make: *** [Makefile:146: all] Error 2
```
Maybe it's only a coincidence.

It's known the libical build can fail in parallel build due to the libical-glib internals (like https://github.com/libical/libical/issues/723), this vala related one was new to me. I've been thinking recently that mabe the dependencies on the files are not that deterministic. I do not know how CMake does it, but if it just waits for a file existence, then the following command can be run before the generator finishes writing to the file (that is what I suspect for the #723), thus if the tools would write to a temporary file and create the result as an "atomic" rename of the file, then it could, theoretically, fix the issues. This is all only a theory of me, most likely a wrong one.

mcrha:
> other commits are RB

What does "RB" stand for here, please? "Look good" might be "LG", I guess.

tintou:
Sorry, Rb is Reviewed-by, so looks good to me :)

tintou:
Yes that's probably a racing issue because vapigen really only use the .gir file

PR #806, opened 2024-11-06
build: Update uninstall.cmake.in

mcrha:
The 'exec_program' command is deprecated in a newer CMake, resulting in the below warning for each uninstalled file.

```
CMake Warning (dev) at uninstall.cmake:17 (exec_program):
  Policy CMP0153 is not set: The exec_program command should not be called.
  Run "cmake --help-policy CMP0153" for policy details.  Use the cmake_policy
  command to set the policy and suppress this warning.

  Use execute_process() instead.
This warning is for project developers.  Use -Wno-dev to suppress it.
```

### DIFF
--- a/uninstall.cmake.in
+++ b/uninstall.cmake.in
@@ -14,10 +14,10 @@ string(REGEX REPLACE "\n" ";" files "${files}")
 foreach(file ${files})
   message(STATUS "Uninstalling $ENV{DESTDIR}${file}")
   if(IS_SYMLINK "$ENV{DESTDIR}${file}" OR EXISTS "$ENV{DESTDIR}${file}")
-    exec_program(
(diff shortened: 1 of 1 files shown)


winterz:
This won't work since we still support CMake versions < 3.28
I'll look at another way to solve the problem.

mcrha:
The top CMakeList.txt currently says `cmake_minimum_required(VERSION 3.20.0)` and the `execute_process` command is available even in 3.1 https://cmake.org/cmake/help/v3.1/command/execute_process.html , thus I guess it should be fine. Or did I overlook anything?

winterz:
> The top CMakeList.txt currently says `cmake_minimum_required(VERSION 3.20.0)` and the `execute_process` command is available even in 3.1 https://cmake.org/cmake/help/v3.1/command/execute_process.html , thus I guess it should be fine. Or did I overlook anything?

yes, you are correct.

PR #807, opened 2024-11-06
icaltime: Fails to convert time to time_t before epoch

mcrha:
The time_t of a time before epoch is a negative number, but the code considered a negative number as an incorrectly entered time.


mcrha:
The 3.0 branch is affected by this as well, thought the patch is slightly different due to other API changes (`icaltime_t` addition). I can open a new merge request against the 3.0 branch is this change is approved.

winterz:
@mcrha thanks. yes please cherry-pick into 3.0

Could you also add to the 3.0 ReleaseNotes.txt "fixed datetimes before the start of the epoch" (or somesuch)

mcrha:
Here you are: https://github.com/libical/libical/pull/809

Issue #808, opened 2024-11-08
Empty build for glib not so empty


dilyanpalauzov:
A build is empty, when typing twice in a row `make` or `ninja`, the second execution does nothing.

When I build libical with GLib binding, typing several times `ninja` does every time execute `ical-glib-src-generator src/libical-glib/tools src/libical-glib/api`.  This should be executed only once, the first time.  Subsequent build should be empty.  In particular, when I build as non-root, on `sudo ninja install` the ownership of the files is changed.

[CMakeCache.txt](https://github.com/user-attachments/files/17674377/CMakeCache.txt)

[build.ninja.txt](https://github.com/user-attachments/files/17674390/build.ninja.txt)

winterz:
please let me know which branch.
the buildsystems are starting to diverge between master and 3.0

dilyanpalauzov:
`master` branch.

dilyanpalauzov:
And `3.0` branch.

dilyanpalauzov:
Yes, I know.  The problem exists in f54ee2fb, does not exist with the previous commit 828405a2, so it was indeed introduced by f54ee2fb.

winterz:
Please test 1e73d13c3fafc24690d93f84bbcc2b6731f07188

if it works I will cherry-pick into 3.0

dilyanpalauzov:
With https://github.com/libical/libical/commit/1e73d13c3fafc24690d93f84bbcc2b6731f07188 I get
```
 CMake Error at src/libical-glib/CMakeLists.txt:156 (add_executable):
   Cannot find source file:

     /home/d/git/libical/build/src/libical-glib/ical-glib-build-check.c

 CMake Error at src/libical-glib/CMakeLists.txt:156 (add_executable):
   No SOURCES given to target: ical-glib-build-check
```
for an out-of-tree/VPATH build)

winterz:
please try with the new patch #813

dilyanpalauzov:
With https://github.com/libical/libical/pull/813 `CMake Error at src/libical-glib/CMakeLists.txt:156 (add_executable):Cannot find source file …` does not appear any more.  But the empty build is not empty.

winterz:
I reverted and cleaned a bit .
please try 2a0a5ba16eeac4d9531d7236d5144c033ea3c6b0

dilyanpalauzov:
Now the build process behaves as expected.

PR #812, opened 2024-11-10
libical-glib/CMakeLists.txt - ical-glib-header target fix

winterz:
Remove ical-glib-build-check.c from the BYPRODUCTS list. The ical-glib-header does NOT create ical-glib-build-check.c

Mistakenly added in commit f54ee2f

fixes: #808


winterz:
never mind. I'll try again later

PR #815, opened 2024-11-13
Added additional byref function that do not support structs as returntype

siam-smeets:
Some languages does not support structs as return type in their C interop to allow those languages to use this external library.

### DIFF
--- a/src/libical/CMakeLists.txt
+++ b/src/libical/CMakeLists.txt
@@ -216,6 +216,7 @@ set(ical_LIB_SRCS
     icallangbind.c
     caldate.c
     astime.h
+    byref.c
 )
 if(LIBICAL_DEVMODE_MEMORY_CONSISTENCY)
   list(APPEND ical_LIB_SRCS ${PROJECT_SOURCE_DIR}/src/test/test-malloc.c)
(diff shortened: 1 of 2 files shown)


tintou:
What language are you referring to? It feels like that this should be included in the language binding instead…

winterz:
I'd like more info , what C version or C standard is required here
does relaxing from C99 to older standards help?

I'm not against this patch, but would like more details and documentation.

siam-smeets:
In this case it is not about the c standard. There are multiple languages that only support pointers but not structures directly as interop because their internal structure is different from C. For us this is for the programming language DataFlex which does support a lot of interop and even C compatible structs for the most part except in return values from a C library.

To circumvent this restriction it is possible in DataFlex and possibly other languages to pass a pointer to a memory layout and fill in that memory space. This is where this "byref" concept comes from. This allows us to use libical with your current structures without much changes etc. It is fine if you guys want a different name for the file/concept or function convention but it would be appreciated if this could be incorporated up-stream.

Hope to have given you enough information.

winterz:
ok.  could you add " There are multiple languages that only support pointers but not structures directly as interop ..."
as a comment at the top of byref.c

and any other documentation you think would be useful in there.

Issue #782, opened 2024-10-07
`BYWEEKNO` inconsistent behavior when built without libicu


minichma:
The following test case succeeds when built with ICU but fails without:

```c
void test_weekno_with_dayno(void)
{
    struct icalrecurrencetype *recur = icalrecurrencetype_new_from_string("FREQ=YEARLY;BYWEEKNO=6;BYDAY=TU;WKST=TH;UNTIL=20210612T000000Z");
    icalrecur_iterator *it = icalrecur_iterator_new(recur, icaltime_from_string("20180206T080001"));
    icalrecurrencetype_unref(recur);

    icaltimetype t;
    
    t = icaltime_from_string("20180213T080001");
    ok("next recurrence is 20180213T080001", icaltime_compare(icalrecur_iterator_next(it), t) == 0);
    t = icaltime_from_string("20190212T080001");
    ok("next recurrence is 20190212T080001", icaltime_compare(icalrecur_iterator_next(it), t) == 0);
    t = icaltime_from_string("20200211T080001");
    ok("next recurrence is 20200211T080001", icaltime_compare(icalrecur_iterator_next(it), t) == 0);
    t = icaltime_from_string("20210209T080001");
    ok("next recurrence is 20210209T080001", icaltime_compare(icalrecur_iterator_next(it), t) == 0);
    
    icalrecur_iterator_free(it);
}
```

In particular the instance `20200211T080001` is missing when run from a build with `HAVE_LIBICU` undefined. It seems that the behavior without ICE is wrong, i.e. `20200211T080001` is a correct recurrence and should be returned.

winterz:
which brings up the issue: should we require libicu now.

note that we have vcpkg support now so Windows people can use that to install libcu pretty easily.

minichma:
Hm, I'm quite happy that there's support for building without libicu, as in our projects we use the lib on a small ARM Cortex M MCU, where we don't have the required resources available.

winterz:
ack. 
no problem. we can keep the libicu dependency optional

ksmurchison:
I will try to give this a look this week, but I'm at a conference through Thu, so no promises.

PR #823, opened 2024-11-22
Fix testcase for github issue782, which had an illegal DTSTART

minichma:
The following testcase was wrong, because `DTSTART` was not in line with the `RRULE` and therefore illegal:

```
# github issue782
RRULE:FREQ=YEARLY;BYWEEKNO=6;BYDAY=TU;WKST=TH;UNTIL=20210612T000000Z
DTSTART:20180206T080001
INSTANCES:20180213T080001,20190212T080001,20200211T080001,20210209T080001
```

This PR fixes the test case by changing `DTSTART` to `20180213T080001`


winterz (approved):

ksmurchison:
That test came from the data that you posted in issue #782 .  Even though its technically invalid to have DTSTART not match the RRULE, clients actually do this all the time.

minichma:
I know, it was my test-case, which is why I feel responsible for fixing it ;-) The problem in #782 would also have appeared with a correct DTSTART, just double-checked (though the error was 'MALFORMED DATA' rather than a wrong instance). The intention of the test was to reproduce the problem in a *legal* case, so its clear, that it reproduced a *real* issue rather than an undefined case. I think we should have separate tests for DTSTART not matching the RRULE. But do we even have a clear definition of how to deal with that case? (I fully agree though, that DTSTART should NOT be included if it doesn't match.) Anyhow, I think the test should be corrected. Do you think the previous version should be preserved as well? I'll be happy to restore it.

Issue #630, opened 2022-11-25
icalcomponent_get_dtend() should treat the DURATION's time part as accurrate value.


minichma:
RFC 5545 specifies how to deal with DURATION properties in case of discontinuities in the time scale (i.e. DST changes). In such cases the date part (days, weeks) should be considered as what they call a nominal value and the time part should be considered as accurate value.

The RFC says (see [here](https://www.rfc-editor.org/rfc/rfc5545#section-3.3.6)):
> The format can represent nominal durations (weeks and days) and accurate durations (hours, minutes, and seconds). ... The duration of a week or a day depends on its position in the calendar.  In the case of discontinuities in the time scale, such as the change from standard time to daylight time and back, the computation of the exact duration requires the subtraction or addition of the change of duration of the discontinuity

If I understand this correctly, then the way we calculate DTEND from DURATION in `icalcomponent_get_dtend()` (and vice versa in `icalcomponent_get_duration()`) is wrong, because we treat both parts of the DURATION property as accurate values while we should treat the date part as nominal value.

This is my understanding in pseudo-code:

_dtend_utc := to_utc(dtstart_local + duration.days) + duration.time_

This is also represented by the following test case:

```c
static void test_icalcomponent_get_dtend_from_duration_single(const char* dtstart, const char* duration, const char* expected_dtend) {

    icalcomponent* comp = icalcomponent_new(ICAL_VEVENT_COMPONENT);

    icaltimetype dtstartt = icaltime_from_string(dtstart);
    dtstartt.zone = icaltimezone_get_builtin_timezone("Europe/Vienna");
    icalcomponent_set_dtstart(comp, dtstartt);

    icalcomponent_set_duration(comp, icaldurationtype_from_string(duration));

    ok("icalcomponent_get_dtend must consider DURATION's time part as accurate.", icaltime_compare(icalcomponent_get_dtend(comp), icaltime_from_string(expected_dtend)) == 0);

    icalcomponent_free(comp);

}

void test_icalcomponent_get_dtend_from_duration(void)
{
    // The test cases refer to the 'Europe/Vienna' time zone, where DST ended at 2022-10-30 3:00.

    // OK
    // Duration doesn't cross DST end.
    test_icalcomponent_get_dtend_from_duration_single("20221029T010000", "PT4H", "20221029T050000");

    // NOT OK
    // Duration crosses DST end. As it's only the time part, it must be considered as accurate duration.
[...]

Issue #633, opened 2022-12-21
Time span calculation of all-day events is broken


christoph-zens:
According to RFC5545, DTEND specifies the non-inclusive end of a VEVENT:

      The "DTSTART" property for a "VEVENT" specifies the inclusive
      start of the event.  For recurring events, it also specifies the
      very first instance in the recurrence set.  The "DTEND" property
      for a "VEVENT" calendar component specifies the non-inclusive end
      of the event.  For cases where a "VEVENT" calendar component
      specifies a "DTSTART" property with a DATE value type but no
      "DTEND" nor "DURATION" property, the event's duration is taken to
      be one day.

An example is given in the RFC, making clear that the end date is the first day on which the event no longer applies:

      The following is an example of the "VEVENT" calendar component
      used to represent a multi-day event scheduled from June 28th, 2007
      to July 8th, 2007 inclusively.  Note that the "DTEND" property is
      set to July 9th, 2007, since the "DTEND" property specifies the
      non-inclusive end of the event.

       BEGIN:VEVENT
       UID:20070423T123432Z-541111@example.com
       DTSTAMP:20070423T123432Z
       DTSTART;VALUE=DATE:20070628
       DTEND;VALUE=DATE:20070709
       SUMMARY:Festival International de Jazz de Montreal
       TRANSP:TRANSPARENT
       END:VEVENT

When an all-day event is specified by DTSTART only, there is code to implement the above statement of assuming a span of one day, with DTEND being set to the day after DTSTART. This is for example done correctly in icalcomponent_foreach_recurrence():

```
/* The end time could be specified as either a DTEND or a DURATION */
/* icalcomponent_get_dtend takes care of these cases. */
dtend = icalcomponent_get_dtend(comp);
if (icaltime_is_null_time(dtend) && icaltime_is_date(dtstart)) {
    /* No DTEND or DURATION and DTSTART is DATE - duration is 1 day */
    struct icaldurationtype dur = icaldurationtype_null_duration();
 
    dur.days = 1;
    dtend = icaltime_add(dtstart, dur);
}

``` 
 With the new DTEND (or the given one, which is assumed to be the day after the last event day), icaltime_span_new() is called:
 
```
/* Now set up the base span for this item, corresponding to the
       base DTSTART and DTEND */
basespan = icaltime_span_new(dtstart, dtend, 1);

```
This function now returns a time span of **48 hours** (-1 second), which is wrong.
[...]

PR #846, opened 2024-12-15
Allow `icalcomponent_foreach_recurrence` to receive `DATE`-only `start` and `end` params.

### LINKED ISSUE #833: icalcomponent_foreach_recurrence() misses recurrences if called with a icaltime with is_date=1
Patch showing the behaviour that this is about (based on release 3.0.18):
```diff
diff --git a/src/libical/icalcomponent.c b/src/libical/icalcomponent.c
index c89f4cfd..4546e605 100644
--- a/src/libical/icalcomponent.c
+++ b/src/libical/icalcomponent.c
@@ -865,6 +865,7 @@ void icalcomponent_foreach_recurrence(icalcomponent *comp,
 
             /* make sure we include any recurrence that ends in timespan */
             icaltime_adjust(&mystart, 0, 0, 0, -(int)(long)dtduration);
+            printf("Adding %ld to %s results in new start of %s\n", -(long)dtduration, icaltime_as_ical_string(start), icaltime_as_ical_string(mystart));
             icalrecur_iterator_set_start(rrule_itr, mystart);
         }
 
diff --git a/src/test/regression.c b/src/test/regression.c
index 0b36000b..3cf91a32 100644
--- a/src/test/regression.c
+++ b/src/test/regression.c
@@ -602,6 +602,43 @@ int test_component_foreach_parameterized(int startOffsSec, int endOffsSec, int e
     return 0;
 }
 
+void test_hack_for_bug_report(void)
+{
+    const char *calStr =
+        "BEGIN:VCALENDAR\n"
+        "BEGIN:VEVENT\n"
+        "DTSTART:20240323T010000Z\n"
+        "DTEND:20240324T200000Z\n"
+        "RRULE:FREQ=WEEKLY\n"
+        "END:VEVENT\n"
+        "END:VCALENDAR\n";
+
+    icalcomponent *calendar = icalparser_parse_string(calStr);
+    icalcomponent *event = icalcomponent_get_first_component(calendar, ICAL_VEVENT_COMPONENT);
+    for (int is_date = 0; is_date < 2; is_date++) {
+        printf("Testing with is_date=%d\n", is_date);
+
+        // 2024-03-24 00:00 UTC to 2024-03-30 00:00 UTC
+        struct icaltimetype t_start = icaltime_from_timet_with_zone(1711238400, is_date, NULL);
+        struct icaltimetype t_end = icaltime_from_timet_with_zone(1711756800, is_date, NULL);
+
+        // Check that this is really exactly midnight.
+        int_is("start hour", t_start.hour, 0);
+        int_is("end hour", t_end.hour, 0);
+        int_is("start minute", t_start.minute, 0);
+        int_is("end minute", t_end.minute, 0);
+        int_is("start second", t_start.second, 0);
+        int_is("end second", t_end.second, 0);
+
+        int found = 0;
+        icalcomponent_foreach_recurrence(event, t_start, t_end, test_component_foreach_callback, &found);
+        int_is("Occurs", found, 1);
+
+        puts("Done with test");
+    }
+    icalcomponent_free(calendar);
+}
+
 void test_component_foreach(void)
 {
     const char *calStr =
@@ -5274,6 +5311,8 @@ int main(int argc, char *argv[])
     test_run("Test icalcomponent_vanew with lastmodified property", test_icalcomponent_with_lastmodified, do_test, do_header);
     test_run("Test attendees", test_attendees, do_test, do_header);
 
+    test_run("hack for bug report", test_hack_for_bug_report, do_test, do_header);
+
     /** OPTIONAL TESTS go here... **/
 
 #if defined(WITH_CXX_BINDINGS)
```
Relevant output when running `bin/regression`:
```
########## hack for bug report                      (80) ##########
Testing with is_date=0
ok 864 - start hour
ok 865 - end hour
ok 866 - start minute
ok 867 - end minute
ok 868 - start second
ok 869 - end second
Adding -154800 to 20240324T000000 results in new start of 20240322T050000
ok 870 - Occurs
Done with test
Testing with is_date=1
ok 871 - start hour
ok 872 - end hour
ok 873 - start minute
ok 874 - end minute
ok 875 - start second
ok 876 - end second
Adding -154800 to 20240324 results in new start of 20240324
not ok 877 - Occurs
# test failed: ""
#          at: src/test/regression.c:635
#      got: 0
# expected: 1
Done with test
```
As you can see, when calling `icalcomponent_foreach_recurrence` with an `icaltime` with `is_date=1`, this code misses a recurrence of the event. This is because the code where I added a `printf` takes the event duration (in this case 43 hours) as a value in seconds and passes it to `icaltime_adjust(&mystart, 0, 0, 0, -43*60*60);`. This function then just does nothing here: https://github.com/libical/libical/blob/80045d9c46006ccdbd8d10349edd607974eb095b/src/libical/icaltime.c#L760-L762
Figuring this out took me quite some time.

Ideas for a fix:
- Make `icalcomponent_foreach_recurrence()` reject `start` or `end` being times (e.g. via some `icalerror_set_errno()` plus an early return)
- Make `icalcomponent_foreach_recurrence()` silently set `.is_date = 0` on its `start` and `end` parameters. Dunno if this really works, but I guess dates should always have their hour, minute, second components set to zero...?
- Somehow normalise the duration in `icalcomponent_foreach_recurrence()`. E.g.: `int one_day = 24*60*60; int days = (dtduration + one_day - 1) / one_day; int seconds = dtduration - days * one_day; icaltime_adjust(&mystart, -days, 0, 0, -seconds);`
- Make `icaltime_adjust()` not ignore its second parameter if it is more than a day in seconds. Dunno what the desired behaviour would be (round up or down?) nor do I doubt that this change in behaviour is a good idea.
- Just leave the behaviour as-is and somehow document this case as unsupported? (Dunno if there is anything else that might have a problem with dates)

minichma:
Fixes #833.

* Allow `icalcomponent_foreach_recurrence` to receive `DATE`-only `start` and `end` params.
* Fix incorrect calculation of an event's duration in case `DTSTART`/`DTEND` are DATE-only.
* Mention limitation that the lib doesn't distinguish between nominal and exact event durations (related to #630).

### DIFF
--- a/ReleaseNotes.txt
+++ b/ReleaseNotes.txt
@@ -98,6 +98,8 @@ Version 3.0.19 (UNRELEASED):
  * Resolved known limitation: Negative values are now also supported for `BYMONTHDAY` and `BYYEARDAY`.
  * Add support for RDATE;VALUE=PERIOD
  * Fix time conversion to time_t for times before epoch
+ * Allow `icalcomponent_foreach_recurrence` to receive `DATE`-only `start` and `end` params.
+ * Fix the calculation of an event's duration if `DTSTART` is a `DATE`-only value.
 
 Version 3.0.18 (31 March 2024):
 -------------------------------
--- a/ReleaseNotes.txt
+++ b/ReleaseNotes.txt
@@ -100,6 +100,7 @@ Version 3.0.19 (UNRELEASED):
  * Fix time conversion to time_t for times before epoch
  * Allow `icalcomponent_foreach_recurrence` to receive `DATE`-only `start` and `end` params.
  * Fix the calculation of an event's duration if `DTSTART` is a `DATE`-only value.
+ * Fix `icaltime_span_new()` - ignore the case where DTEND is unset and require it to be set by the caller instead.
 
 Version 3.0.18 (31 March 2024):
 -------------------------------
--- a/docs/KnownLimitations.md
+++ b/docs/KnownLimitations.md
@@ -17,3 +17,14 @@ Contributions to help us fix these limitations are welcome.
 * FREQ=YEARLY, BYWEEKNO can't be combined with BYYEARDAY, BYMONTH or BYMONTHDAY
 
      ref: <https://github.com/libical/libical/blob/cfd401b9d043214395888de1d9daf52263e3245b/src/libical/icalrecur.c#L2928>
+
+### Recurrence Rule sections 3.8.5.3, DURATION 3.3.6
+
+* The lib (e.g. `icalcomponent_foreach_recurrence()`, `icalcomponent_get_dtend()`) does not
+  differentiate between nominal and exact durations. According to the RFC, when a component's
+  duration is specified using the `DURATION` property rather than `DTEND`, the nominal duration
+  must be used to calculate each recurrence's duration. However, the current implementation always
+  uses the exact duration, which can cause discrepancies if a recurrence spans a daylight saving
+  time transition.
+
+     ref: <https://github.com/libical/libical/issues/630>
--- a/src/libical/icalcomponent.c
+++ b/src/libical/icalcomponent.c
@@ -840,6 +840,21 @@ static bool icalcomponent_is_busy(icalcomponent *comp)
     return (ret);
 }
 
+static struct icaltimetype icaltime_with_time(const struct icaltimetype t, int hour, int minutes, int seconds)
+{
+    struct icaltimetype ret = t;
+    ret.hour = hour;
+    ret.minute = minutes;
+    ret.second = seconds;
+    ret.is_date = 0;
+    return ret;
+}
+
+static struct icaltimetype icaltime_at_midnight(const struct icaltimetype t)
+{
+    return icaltime_with_time(t, 0, 0, 0);
+}
+
 void icalcomponent_foreach_recurrence(icalcomponent *comp,
                                       struct icaltimetype start,
                                       struct icaltimetype end,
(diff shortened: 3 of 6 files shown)


winterz (approved):
@psychon please test

once this lands I will cherrypick to the 3.0 branch for the upcoming 3.0.19 release

minichma:
@winterz As discussed in #633 I modified `icaltime_span_new()` in that I removed the special handling of the case that `end` is unset. Within libical this case doesn't occur, which probably is the reason for the problem not popping up more frequently. However, in v3 this change might be breaking for certain users. On the other hand, as it was broken anyways, I wouldn't expect real life problems. What do you think? Alternatively we could just leave this aspect as it was, as it doesn't do any harm either.

winterz:
> @winterz As discussed in #633 I modified `icaltime_span_new()` in that I removed the special handling of the case that `end` is unset. Within libical this case doesn't occur, which probably is the reason for the problem not popping up more frequently. However, in v3 this change might be breaking for certain users. On the other hand, as it was broken anyways, I wouldn't expect real life problems. What do you think? Alternatively we could just leave this aspect as it was, as it doesn't do any harm either.
> 
> [Edit] ... or ignore the last commit when cherry-picking into v3.

we need to deal with any pain now (in the new v3 with this added) or later in 4.0
may as well go for it now with v3

winterz:
I'm happy if you do the merge
can you also do the cherry-pick into 3.0?

minichma:
Certainly. Done.

PR #851, opened 2024-12-24
Clean up the swapping macro handling for BSD's

### LINKED ISSUE #426: bswap_32 wrong on FreeBSD
Upgrading 3.0.4 to 3.0.8 on FreeBSD:

```
ld: error: lib/libical.so.3.0.8: undefined reference to bswap_32
```

Looking at 3.0.8, this.. very nested sequence of ifdefs:

```c
#if defined(HAVE_ENDIAN_H)
#include <endian.h>
#else
#if defined(HAVE_SYS_ENDIAN_H)
#include <sys/endian.h>
#if defined(bswap32)
#define bswap_32 bswap32
#else
#define bswap_32 swap32
#endif
#endif
#endif
```

fails to consider the case where `<endian.h>` is basically just an alias to `<sys/endian.h>` which defines `bswap32` as a macro.

brad0:
Make sure the 64-bit macro is also defined. A build out of the
box would crash with copycluster.

Closes #426

### DIFF
--- a/src/libical/icaltz-util.c
+++ b/src/libical/icaltz-util.c
@@ -20,24 +20,19 @@
 #include <stdlib.h>
 #include <limits.h>
 
+#if defined(sun) && defined(__SVR4)
+#include <sys/types.h>
+#include <sys/byteorder.h>
+#else
 #if defined(HAVE_BYTESWAP_H)
 #include <byteswap.h>
 #endif
 #if defined(HAVE_ENDIAN_H)
 #include <endian.h>
-#else
+#endif
 #if defined(HAVE_SYS_ENDIAN_H)
 #include <sys/endian.h>
-#if defined(bswap32)
-#define bswap_32 bswap32
-#else
-#define bswap_32 swap32
 #endif
-#endif
-#endif
-
-#if defined(__OpenBSD__) && !defined(bswap_32)
-#define bswap_32 swap32
 #endif
 
 #if defined(_MSC_VER)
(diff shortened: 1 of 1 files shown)


src/libical/icaltz-util.c
```diff
@@ -38,6 +38,7 @@
 
 #if defined(__OpenBSD__) && !defined(bswap_32)
 #define bswap_32 swap32
+#define bswap_64 swap64
```
winterz, line 41: would it be better to have:
```
#if defined(__OpenBSD__) && !defined(bswap_64)
#define bswap_64 swap64
#endif
```
?

winterz (commented):
this should be cherry-picked into the 3.0 branch too, I think

src/libical/icaltz-util.c
brad0: I don't think so. The check is just to determine which BSD is being used. Some use bswap macros and some use swap.

brad0:
I took another look at this and there was room for cleaning things up. The OpenBSD specific bit can be removed. I also noticed FreeBSD has a patch which this will fix.

brad0:
> this should be cherry-picked into the 3.0 branch too, I think

Yes, please.

src/libical/icaltz-util.c
```diff
@@ -63,6 +58,17 @@
 #include <io.h>
 #endif
 
+#if defined(HAVE_ENDIAN_HH) || defined(HAVE_SYS_ENDIAN_H)
```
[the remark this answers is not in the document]
brad0: Yes, that is what I meant. Corrected.

winterz:
can you make sure PR #426 is also properly handled in your patch?

brad0:
> can you make sure PR #426 is also properly handled in your patch?

Part of why I rearranged things was to deal with the symptom described in the issue.

Issue #849, opened 2024-12-17
LIBICAL_ENABLE_64BIT_ICALTIME_T Improve


winterz:
Use CMake to determine if the system supports __time64_t

if LIBICAL_ENABLE_64BIT_ICALTIME_T set and __time64_t is not supported then fail with a nice message.

IOW: 
```
  if(MSVC)
    set(ICAL_ICALTIME_T_TYPE "__time64_t")
  else()
    message(FATAL_ERROR)
 endif()
```

becomes
```
  if(HAVE___TIME64_T)
    set(ICAL_ICALTIME_T_TYPE "__time64_t")
  else()
    message(FATAL_ERROR)
 endif()
```

adjust documentation and FATAL_ERROR messages accordingly

see #573 and #701

Issue #852, opened 2024-12-27
libical testcase now fails to build on hppa and powerpc on Debian


babelouest:
Hello, 

As the libical maintainer for Debian, a bug emerged in the last release: https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=1091461

Lately, Debian switched the `time_t` length to 64 bits on 32 bits architecture. This change may have caused an error when building test cases on hppa and PowerPC platforms: 

https://buildd.debian.org/status/fetch.php?pkg=libical3&arch=hppa&ver=3.0.19-1&stamp=1735155399&raw=0

```
62: Test command: /usr/bin/python3 "-I" "/build/reproducible-path/libical3-3.0.19/src/test/libical-glib/recurrence.py"
62: Working Directory: /build/reproducible-path/libical3-3.0.19/obj-hppa-linux-gnu/bin
62: Environment variables:
62:  GI_TYPELIB_PATH=/build/reproducible-path/libical3-3.0.19/obj-hppa-linux-gnu/src/libical-glib
62:  ZONEINFO_DIRECTORY=/build/reproducible-path/libical3-3.0.19/zoneinfo
62:  LD_LIBRARY_PATH=/build/reproducible-path/libical3-3.0.19/obj-hppa-linux-gnu/lib
62: Test timeout computed to be: 10000000
62: Traceback (most recent call last):
62:   File "/build/reproducible-path/libical3-3.0.19/src/test/libical-glib/recurrence.py", line 239, in <module>
62:     assert(result[i] == start + i*secs_per_day)
62:            ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
62: AssertionError
62/64 Test #62: libical-glib-recurrence ..............***Failed    1.07 sec
Traceback (most recent call last):
  File "/build/reproducible-path/libical3-3.0.19/src/test/libical-glib/recurrence.py", line 239, in <module>
    assert(result[i] == start + i*secs_per_day)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
AssertionError
```

According to the original issue author

> I think the problem is in src/Net-ICal-Libical/netical.i:
> In this define, "start" and "array" should be "unsigned long long" or "uint64_t":
int icalrecur_expand_recurrence(const char* rule, int start,
                                int count, int *array);
> Note that int or long can be 64-bit on 64-bit platforms, but on 32-bit platforms you need
[...]

PR #856, opened 2024-12-31
Buildsystem - better handling 64-bit __time64

winterz:
fixes: #849


winterz:
we still have the problem with hard-coded `time_t` in the glib bindings as seen in #852

minichma (approved):
LGTM. Wish you a great 2025 btw :-)

PR #857, opened 2024-12-31
Fix Warnings in glib

winterz:
```
i-cal-component.c:1602: Warning: ICalGLib: invalid "closure" annotation: only valid on callback parameters
i-cal-component.c:1635: Warning: ICalGLib: invalid "closure" annotation: only valid on callback parameters
i-cal-mime.c:20: Warning: ICalGLib: invalid "closure" annotation: only valid on callback parameters
i-cal-parser.c:155: Warning: ICalGLib: invalid "closure" annotation: only valid on callback parameters
i-cal-parser.c:200: Warning: ICalGLib: invalid "closure" annotation: only valid on callback parameters
```

### DIFF
--- a/src/libical-glib/api/i-cal-array.xml
+++ b/src/libical-glib/api/i-cal-array.xml
@@ -51,7 +51,7 @@
     </method>
     <method name="i_cal_array_sort" corresponds="CUSTOM" annotation="skip" kind="others" since="1.0">
         <parameter type="ICalArray *" name="array" comment="The #ICalArray to be sorted"/>
-        <parameter type="gint (*compare)" name="(const void *, const void *)" annotation="scope call" comment="FULL: @compare: (scope call): The compare function"/>
+        <parameter type="gint (*compare)" name="(const void *, const void *)" comment="FULL: @compare: The compare function"/>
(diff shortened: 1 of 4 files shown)


winterz:
it would be nice if we could error on these types of Warnings (similar to how we use -Werror when compiling)

winterz:
CC: @mcrha

mcrha:
It seems GObject introspection compiler does not have any `--error` or similar argument (it has only few arguments, none for "warnings as error").

These annotations tell the GObject introspection what to do with the arguments. The full list can be found [here](https://gi.readthedocs.io/en/latest/annotations/giannotations.html).

I guess, and only guess, the removal of these is safe. For the completeness:
* The `(scope call)` is the default and the `i_cal_array_sort` is a synchronous function. You might remove also ` annotation="scope call"` from the line you modified, to avoid confusion.
* The `closure`-s are supposed to be on the function, not on the argument (I always misplace it, I'm sorry), thus instead of `closure callback` on the `user_data` argument it should be `closure user_data` on the `callback` argument.

winterz:
new version with @mcrha suggestions ready for review

winterz:
thanks for your review
FTR: I have gobject-introspection-1.82.0-1 from Fedora41

PR #866, opened 2025-01-09
Add CMake option for building in "Developer Mode"

winterz:
Developer Mode:
 * compiles with '-Werror' (/WX with MSVC) non-developer-mode by default doesn't use '-Werror'
 * always enables the test suite
 * builds in debug mode by default
 * other things as we add them...


winterz:
Dear developers: once this lands please start using the -DLIBICAL_DEVMOD=True configuration

PR #868, opened 2025-01-15
WIP - Build with Compiler Options Hardening Guide for C and C++

winterz:
Follow the "Compiler Options Hardening Guide for C and C++": <https://best.openssf.org/Compiler-Hardening-Guides/Compiler-Options-Hardening-Guide-for-C-and-C++

The CMake provided from:
 https://www.stevenengelhardt.com/2024/11/12/cmake-implementation-of-openssf-compiler-hardening-options/


winterz:
closing this one.  
I'll make a new patch that doesn't use -Wconversion or -Wsign-conversion as those require a huge number of changes.

PR #872, opened 2025-01-20
Build with gcc/clang -Wconversion option

winterz:


winterz:
closing this.
too painful and dangerous

PR #873, opened 2025-01-21
Add external iterators for properties and parameters

rsto:
In addition to the already existing compiter iterator, this adds external iterators for properties and parameters.

In a separate a drive-by commit, this also adds error checking for some of the compiter functions, which were at risk at dereferencing a NULL pointer.


rsto:
I've converted this to Draft until I'm done with some fine-tuning. I'll ping you when this is ready for review.

winterz:
Thanks looks great.
we do need to add glib bindings for the new public functions some time.
I'll create an issue for that

winterz:
see #874

PR #878, opened 2025-01-25
Coding style - expand line width up to 120 characters

winterz:


winterz:
meh. not so nice.
closing.

PR #881, opened 2025-01-28
icaltypes.c: fix bug ignoring non-empty REQUEST-STATUS extdata

rsto:
Due to an erroneous conditional, a non-empty 'extdata' part of a REQUEST-STATUS property value did not get parsed into the 'debug' field of the icalreqstattype struct type.

It also fixes the docstring in the header that mistakenly stated that the 'debug' field value points into the original input string. Instead, it points into the ring buffer memory.

### DIFF
--- a/src/libical/icaltypes.c
+++ b/src/libical/icaltypes.c
@@ -140,7 +140,7 @@ struct icalreqstattype icalreqstattype_from_string(const char *str)
      */
 
     p2 = strchr(p1 + 1, ';');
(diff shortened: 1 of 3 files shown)


rsto:
I ran across this when I did some testing. I will create another pull request for more hefty edits to the REQUEST-STATUS property parser, but the current PR deliberately is scoped just to a bug fix.

winterz:
thanks

PR #882, opened 2025-01-29
libical-glib: Add iterators for properties and parameters

tintou:
This is a follow-up for commit 1b91c4e34d92c3254fa48ee00b92799997b747ef

Closes: #874


winterz (approved):
maybe you can also add some tests

tintou:
Okay, I've added some naive tests

PR #884, opened 2025-02-04
Split the pkg-config files per library

tintou:
This allows to only link against the minimal set of libraries that we actually use.


winterz (approved):
good idea

PR #885, opened 2025-02-11
libical-glib/cmake: fix environment variable conditionals

K900:
Those could not work as written previously for two reasons:
- `if(DEFINED FOO)` only checks cmake variables, not environment ones
- `if($FOO)` checks for magic truthy values, not any non-empty string, and can also double-expand.

Fix both by using `ENV{FOO}` consistently and explicitly checking for non-emptiness.


winterz:
thanks. I'll cherry-pick to the 3.0 branch

Issue #886, opened 2025-02-11
Broken error reporting for END before BEGIN


psychon:
Hi there,

I have a test case where a random `END:VEVENT` was inserted in the middle of an ical file (in case it matters: right before the `BEGIN:VEVENT` for the first vevent).

With libical 3.0.18, this ends up being indicated through `icalerror_warn()` and is printed to stderr.

With libical 3.0.19, this warning disappeared.

This is because of this change: https://github.com/libical/libical/commit/45188d4a05086ea7b1103617370b5c6af1844aa1#diff-b3128d6a49720c57440111b67b5724f2188a907099daa3515107131deebd5496

The call to `icalerror_warn()` was replaced with some non-sensical call to `icalerror_set_error_state()` that does nothing (these lines are a no-op; they just the current error state to the value it already has). This is completely the wrong function. `icalerror_set_error_state()` is about what should happen if such an error occurs (is it fatal or not?). This is not about actually reporting such an error.

I am not sure, but if I had to guess, I would say that this was meant to be a call to `icalerror_set_errno()`. Alternatively, the rest of the code in this file seems to use `insert_error()` to add a nice-ish error message to the component and then sets `parser->state = ICALPARSER_ERROR`.

winterz:
should be fixed now with 41199f80516ff045f9d2dd39597c7ee6b5269be6

PR #888, opened 2025-02-12
icalparser.c - restore using icalerror_warn() if END is before BEGIN

winterz:
Put back the icalerror_warn() if the parser detects that the END preceeds any BEGIN datetime. Looks like an "ooops" I mistakenly committed when working the fuzzer fixes.

fixes: #886

### DIFF
--- a/src/libical/icalparser.c
+++ b/src/libical/icalparser.c
@@ -834,8 +834,7 @@ icalcomponent *icalparser_add_line(icalparser *parser, char *line)
 
         if (parser->level < 0) {
             // Encountered an END before any BEGIN, this must be invalid data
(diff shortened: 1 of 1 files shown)


psychon:
Thanks for the PR. That is totally fine with me.

> Looks like an "ooops" I mistakenly committed when working the fuzzer fixes.

Just a quick note: I doubt that this was an ooops. The commit message in https://github.com/libical/libical/commit/45188d4a05086ea7b1103617370b5c6af1844aa1#diff-b3128d6a49720c57440111b67b5724f2188a907099daa3515107131deebd5496 mentions some intention (which the patch however did not actually implement, I fear):

> icalparser.c - icalparser_add_line(): END is encountered before the BEGIN becomes an error, not just a warning

PR #891, opened 2025-02-20
libicalvcard: ignore unescaped backslash at end of value

rsto:
Fixes a bug where an unescaped backslash at the end of a value caused the vCard parser attempt to free the multi-valued TEXT buffer, even if that had not been allocated.

Rather than rejecting a TEXT property with a trailing backslash as invalid, the parser now ignores that backslash. This matches what icalvalue_decode_ical_string does for iCalendar values.


ksmurchison (approved):
We have similar code already in production at Fastmail

PR #903, opened 2025-04-24
icalrecur_test: Add test case of the form `FREQ=WEEKLY;INTERVAL=2;WKST=WE;BYDAY=MO,WE,FR`

minichma:
Just another test case related to the [RFC 5545 section 3.3.10 Recurrence Rule](https://www.rfc-editor.org/rfc/rfc5545.html#section-3.3.10), which is saying about `WKST`:

> ... This is significant when a WEEKLY "RRULE" has an interval greater than 1, and a BYDAY rule part is specified.


dilyanpalauzov:
Just for my understanding (not related to the change) what does `WKST=` actually do?  I do not understand on this RFC 5545.  How would the repetitions be different, if `WKST=` had a different value for `FREQ=WEEKLY;INTERVAL=2;BYDAY=...`?  Why is it relevant here when the work week (WKST) starts?

> The WKST rule part specifies the day on which the workweek starts. Valid values are MO, TU, WE, TH, FR, SA, and SU.  This is significant when a WEEKLY "RRULE" has an interval greater than 1, and a BYDAY rule part is specified.  This is also significant when in a YEARLY "RRULE" when a BYWEEKNO rule part is specified.  The default value is MO.

minichma:
It defines, which week a day belongs to. If `FREQ=WEEKLY` and `INTERVAL=1`, this doesn't matter, as every week is considered anyways. But if the interval is different, e.g. `INTERVAL=2`, then the question is, which week a day belongs to. If e.g. WKST=WE and DTSTART is on a Monday, then the following Wednesday already belongs to the next week and wouldn't be returned.

More importantly it's relevant for determining the `WEEKNO`. E.g. week 1 is the first having at least 4d within the given year. This obviously depends on `WKST`.

PR #904, opened 2025-04-30
Fix warnings in C source found by compiling with gcc -fanalyzer

winterz:
TODO: fixes for analyzer warnings in C++ source


winterz:
CC: @mcrha

src/libical-glib/tools/generator.c
```diff
@@ -2065,6 +2072,7 @@ gchar *get_source_run_time_checkers(Method *method, const gchar *nameSpace)
 
         if (parameter && parameter->type && parameter->type[strlen(parameter->type) - 1] == '*') {
             trueType = get_true_type(parameter->type);
+            g_return_val_if_fail(trueType != NULL, NULL);
```
mcrha, line 2075: This may yell about leaked `buffer` variable, I guess in the Coverity Scan, maybe?

The `g_return_val_if_fail` can be "turned off" during compilation (making them no-op), they are mostly used for the coder's errors, not for the input errors.

winterz: right. thanks for finding that.
fix coming up

PR #907, opened 2025-05-05
src/libical-glib/tools/generator.c - fix the check for empty trueType

winterz:
In get_source_run_time_checkers() just continue the parameter loop if the current parameter trueType is empty.


winterz:
CC: @mcrha

src/libical-glib/tools/generator.c
```diff
@@ -2072,7 +2072,9 @@ gchar *get_source_run_time_checkers(Method *method, const gchar *nameSpace)
 
         if (parameter && parameter->type && parameter->type[strlen(parameter->type) - 1] == '*') {
             trueType = get_true_type(parameter->type);
-            g_return_val_if_fail(trueType != NULL, NULL);
+            if (!trueType) {
+                continue;
```
mcrha, line 2076: it seems it's expected the "true type" is always present in the code, then I'd print an error on the console if it's not provided, rather than silently skip it. Something like the below may do the trick:
```c
   if (!trueType) {
      fprintf(stderr, "Unknown type '%s' in method %s\n", parameter->type, method->name);
      continue;
   }
```

mcrha: and looking into the `get_true_type()` implementation it can return `NULL` only if the passed in `type` is a nonsense (`NULL` or empty), thus the non-NULL return of the `get_true_type()` check is kind of fatal, when the function returns the `NULL`.

mcrha:
Looks good to me, thank you.

PR #910, opened 2025-05-06
CMakeLists.txt: ignore out-of-source build folder

yegorich:
Create a `.gitignore` file inside a build folder. Thus, this folder will be ignored by git and, hence, no entry in the root `.gitignore` is required.

For more information, see this post:
https://www.scivision.dev/cmake-auto-gitignore-build-dir/


winterz:
cute.
thanks

PR #911, opened 2025-05-06
Update libical-glib versioning

winterz:


mcrha:
Err, you asked me to write this, but I completely forgot of it, being distracted by other work. I'm sorry.

There are also `ICal-3.0.gir` and `ICal-3.0.typelib` files.

winterz:
> Err, you asked me to write this, but I completely forgot of it, being distracted by other work. I'm sorry.
> 
no problem.

> There are also `ICal-3.0.gir` and `ICal-3.0.typelib` files.
should be fixed with my latest force push

src/test/libical-glib/array.py
```diff
@@ -16,7 +16,7 @@
 
 import gi
 
-gi.require_version('ICalGLib', '3.0')
+gi.require_version('ICalGLib', '4.0')
```
yegorich, line 19: Will it make sense to also use `LIBICAL_LIB_MAJOR_SOVERSION` for this dependency?

mcrha: only if the test files are "configured" as the config.h or similar files, which is not feasible considering it is a test file, not a build file.

mcrha:
The change looks good to me. I did not check other usages of the `CMAKE_PROJECT_VERSION_MAJOR`, if any.

src/test/libical-glib/array.py
yegorich: OK. I see

winterz: > Will it make sense to also use `LIBICAL_LIB_MAJOR_SOVERSION` for this dependency?

I thought about that.
It would be nice not to change all the test `.py` files by-hand every time we update the major version.
then we need to generate them at cmake stage, which should be possible.
in a separate patch.

mcrha: You do not change the version that often to it be any concern, no? The tests will fail soon when the project/soname version is changed, but the python tests not updated.

winterz: > You do not change the version that often to it be any concern, no? The tests will fail soon when the project/soname version is changed, but the python tests not updated.

yes, that's true.
it would be "busy work" if someone is interested.

PR #921, opened 2025-05-30
.github/workflows/build.yml - add the "latest" runners

winterz:


winterz:
I waited an hour and the runners never started.
@tintou can you help?

Issue #243, opened 2016-09-04
Handle multi-valued parameters (namely MEMBER) properly


ksmurchison:
Currently the way parameter values are parsed, a list of MEMBER values is clipped after the first value because the parser sees a quoted string as assumes that its the entire value of the property

PR #928, opened 2025-06-26
Support multi-valued iCalendar parameters

rsto:
This adds support for multi-valued iCalendar parameters. It fixes https://github.com/libical/libical/issues/243

This is working code, but I will leave it as draft until some more experimentation. It supports multiple values for the following parameters:
- DELEGATED-FROM (RFC 5545)
- DELEGATED-TO (RFC 5545)
- MEMBER (RFC 5545)
- DISPLAY (RFC 7986)
- FEATURE (RFC 7986)

Currently, the libical parser discards all but the first parameter value for the DELEGATED-FROM, DELEGATED-TO and MEMBER parameters. This is because the parser ignores any value that follows a quoted-string parameter value. For DISPLAY, it parses all comma-separated enum values but stores them as a single X-parameter value.

Consequently, libical currently does not allow to process multiple values of these parameters or preserve them when rendering the in-memory representation back to iCalendar.

This patch fixes that by reusing the multi-value parameter support that already was available in libicalvcard.

This change is not backwards-compatible, neither for libical or libicalvcard. The changes were made to help migrating existing codebases:

- The `get_foo` function now returns a list of values. A new `get_foo_nth` function allows to access the nth parameter value, a new `get_foo_size` function indicates the length of the parameter value list.
- The `set_foo` function now takes a list of values.
- The `new_foo` constructors take a single value as argument, as this is compatible with libical and likely to be the common usage pattern. To construct a parameter with a list of values, a `new_foo_list` constructor is defined.

The C++ compatibility layer is updated but the glib integration currently is not.

### DIFF
--- a/config.h.cmake
+++ b/config.h.cmake
@@ -466,7 +466,10 @@ typedef int IO_SIZE_T;
 typedef unsigned int IO_SSIZE_T;
 #else
 typedef size_t IO_SIZE_T;
+#cmakedefine HAVE_SSIZE_T
+#if !defined(HAVE_SSIZE_T)
 typedef ssize_t IO_SSIZE_T;
+#endif
 #endif
 
 #if defined(_MSC_VER)
--- a/config.h.cmake
+++ b/config.h.cmake
@@ -466,10 +466,7 @@ typedef int IO_SIZE_T;
 typedef unsigned int IO_SSIZE_T;
 #else
 typedef size_t IO_SIZE_T;
-#cmakedefine HAVE_SSIZE_T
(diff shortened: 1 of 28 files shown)


rsto:
@ksmurchison @winterz before I spend time fixing the CI failures, what's your take on this change?

ksmurchison:
> @ksmurchison @winterz before I spend time fixing the CI failures, what's your take on this change?

LGTM.  Will approve once CI passes

winterz:
I copied this to a work branch (work/winterz/libical_multivalue_params) so we can work on this together rather than being in a fork

PR #929, opened 2025-06-28
Support multi-valued iCalendar parameters

winterz:
This adds support for multi-valued iCalendar parameters. It fixes https://github.com/libical/libical/issues/243

It supports multiple values for the following parameters:

    DELEGATED-FROM (RFC 5545)
    DELEGATED-TO (RFC 5545)
    MEMBER (RFC 5545)
    DISPLAY (RFC 7986)
    FEATURE (RFC 7986)

Currently, the libical parser discards all but the first parameter value for the DELEGATED-FROM, DELEGATED-TO and MEMBER parameters. This is because the parser ignores any value that follows a quoted-string parameter value. For DISPLAY, it parses all comma-separated enum values but stores them as a single X-parameter value.

Consequently, libical currently does not allow to process multiple values of these parameters or preserve them when rendering the in-memory representation back to iCalendar.

This patch fixes that by reusing the multi-value parameter support that already was available in libicalvcard.

This change is not backwards-compatible, neither for libical or libicalvcard.
* The `get_foo` function now returns a list of values. A new `get_foo_nth` function allows to access the nth parameter value, a new `get_foo_size` function indicates the length of the parameter value list.
* The` set_foo` function now takes a list of values.
*  The`new_foo` constructors take a single value as argument, as this is compatible with libical and likely to be the common usage pattern. To construct a parameter with a list of values, a `new_foo_list` constructor is defined.


winterz:
@rsto can you use a different type in the  headers instead of `ssize_t` as MacOS doesn't really know that typedef.

is `long` good enough?
else perhaps `int64_t`

rsto:
@winterz we could make the build work for `ssize_t` but I propose we get rid of it. I tried pushing a proposal to this branch but my Github account does not have the permissions.

On my own branch, I pushed commit https://github.com/libical/libical/pull/928/commits/4a6ec2a4c991955e90509c5320088ccadcb0b41e which refactors the uses of `ssize_t` to `size_t`. This is a breaking change for any code using the `vcardstrarray` and `vcardenumarray` types, but on the other hand the current `ssize_t`-based API implementation is broken anyway: e.g. calling `vcardstrarray_remove_element_at` with a negative index did not, as one would assume, remove the nth-from-the-last element, but instead caused a memory corruption! I think it's fair enough to just get rid of these negative indexes in the API.

Should we actually want to keep using `ssize_t` then the following snippet in the headers is enough (I verified this with CI):
```
#if defined(_MSC_VER)
#include <BaseTsd.h>
typedef SSIZE_T ssize_t;
#else
#include <sys/types.h>
#endif
```

Could you please allow my Github account to push to this branch, or cherry-pick the ongoing work from my branch?

winterz:
I just sent you an invitation to join this project as a developer

rsto:
@winterz thanks, I pushed my changes and CI passes. I ran the regression tests on valgrind locally and it ran OK. Could you please rerun AddressSaninitizer in case it's not part of CI?

rsto:
I fixed the failing AddressSanitizer test and also make the default increment size for icalarray a preprocessor definition. I had added this to CMakeList.txt but factored it into icalarray.c itself, I don't know what you prefer but found defining it in icalarray.c to be more clear for now.

I also started writing API documentation in the new header files but while doing so realized that the `array_remove` functions do not check for duplicate entries in the array, but they should. I will spend some more time fixing that, too.

rsto:
BTW I came to realize that I already had supported a default increment size in an earlier commit in that branch, but it had defaulted to `1` which really is bad for performance and only was for testing.

src/libical/icalenumarray.h
```diff
@@ -0,0 +1,55 @@
...
+    const char *xvalue;
+} icalenumarray_element;
+
+#define icalenumarray_new(increment_size) \
+    icalarray_new(sizeof(icalenumarray_element), increment_size)
+
+#define icalenumarray_element_at(array, position) \
+    ((const icalenumarray_element *)icalarray_element_at(array, position))
+
+#define icalenumarray_size(array) ((array)->num_elements)
```
winterz, line 33: we need to consider a dereference of a null pointer situation.
maybe:
`#define icalenumarray_size(array) (if(array) ? (array)->num_elements) : -1)` ?

same for the `vcardenumarray_size(array)` macro

rsto: I'm all for being defensive against NULL pointer arguments, I just kept it as-is because none of the array functions, including icalarray are safe anyway. I'll submit a patch for that in this PR, too.

rsto:
@winterz I just noticed that the derived "new" constructors for multi-valued enum parameters aren't providing the API that they were supposed to -> they take an enumarray_element value but should take a numeric enum. That's for backwards compatibility and I just noticed this while upgrading our own code. I'll push a change to this PR soon.

winterz:
compiling with gcc analyzer:

```
/home/winterz/projects/libical/libical/src/libical/icalstrarray.c:45:14: warning: use of NULL where non-null expected [CWE-476] [-Wanalyzer-null-argument]
```

winterz:
next I will test build against the KDE kcalendarcore library and adjust the migration document, as necessary

PR #932, opened 2025-07-05
Set LINKER_LANGUAGE for doesnothing

alyssais:
This is needed for the same reason as icalrecurtest in the linked PR, as another C program linking with a C++ library.

Since the previous PR, I've learned a bit more about the situation here. As you can read in https://github.com/pkgconf/pkgconf/issues/377, it's basically an unsolved problem to use dependencies that might or might not have C++ code in them, but it's at least possible to do the right thing within a single project like this.

### DIFF
--- a/examples/CMakeLists.txt
+++ b/examples/CMakeLists.txt
@@ -28,6 +28,10 @@ add_dependencies(
   icalss
   icalvcal
 )
+if(WITH_CXX_BINDINGS)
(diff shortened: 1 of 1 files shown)


winterz:
maybe we should check at the top-level CMakeLists.txt for `WITH_CXX` and set the `LINKER_LANGUAGE` for all targets.  and then be done with it  for all current and future targets

Issue #794, opened 2024-10-23
BYWEEKNO recurrence issue


minichma:
The following cases related to `BYWEEKNO` fail:

```
RRULE:FREQ=YEARLY;BYWEEKNO=1,2;UNTIL=20170101T000000Z
DTSTART:20130108T000000
Expected INSTANCES:20130108T000000,20131231T000000,20140107T000000,20141230T000000,20150106T000000,20160105T000000,20160112T000000
Actual   INSTANCES:20130115T000000,20130122T000000,20140114T000000,20140121T000000,20150113T000000,20150120T000000,20160112T000000,20160119T000000

RRULE:FREQ=YEARLY;BYWEEKNO=-1,-2
DTSTART:20131224T000000
Expected INSTANCES:20131224T000000
Actual   INSTANCES: *** MALFORMEDDATA: An input string was not correctly formed or a component has missing or extra properties
```

(the expected instances are based on existing test cases; didn't check whether they are correct)

Reproduced on Ubuntu with ICU and Windows without ICU.

minichma:
see #794

winterz:
@CMendia would you be willing to work on interpretation #1 to solve this one?

PR #941, opened 2025-08-05
Fix BYWEEKNO handling when BYDAY is not present

CMendia:
The bitmask for BYWEEKNO is incorrectly generated when BYDAY is not present. Tests did not catch this because it was only tested once, with an input that happened to produce the correct output. This changes BYWEEKNO handling without BYDAY to resemble handling with BYDAY.


minichma:
@CMendia Very nice work, appreciate you're approaching the BYWEEKNO topic. I assume your work is related to #794, so please note the warning block I just added to the issue description, i.e. that the expected behavior is not fully clear, at least not from my point of view. I documented my view there, following what seems to have been the chosen approach so far, but I might be fully wrong (you're more than welcome to contribute your point of view on these topics).

Please also note the additional test case I added, which shows one of the border cases related to BYWEEKNO (and which may or may not be correct):

```ics
RRULE:FREQ=YEARLY;BYWEEKNO=-1,-2;COUNT=6
DTSTART:20270102
Expected INSTANCES:20270102,20271225,20280101,20281223,20281230
Actual INSTANCES:20271225,20280101,20281223,20281230,20291222,20291229
```

PR #943, opened 2025-08-05
Add Java bindings

winterz:


abaker:
Can you try this out?

abaker:
The unit tests seem to be passing on the MacOS runner, and they pass locally for me on MacOS. It seems like its not finding the DLL on windows, and I'm not sure why its crashing on Ubuntu 😢

winterz:
on Windows, perhaps we need to add the equivalent of JAVA_HOME/bin to PATH

in src/java/CMakeLists.txt make something like:
```
if(WIN32)
set_tests_properties(
      libical_java_test
      PROPERTIES
        ENVIRONMENT
          "PATH=JAVA_HOME/bin"  #whatever
       TIMEOUT 60
    )
endif()
```

winterz:
> looking great.
> 
> all that remains is adding copyright+license info to the top of `cmake/modules/FindJUnit.cmake` and `src/java/test/LibIcalTest.java`

did you write FindJUnit.cmake or borrow it from another project?  so I know how to copyright and license

do you want copyright on LibIcalTest.java or do you want to give copyright to the libical developers?
I'll need your name and email if you want copyright

winterz:
@abaker 
I notice the libical_jni.so library is installed in <install_prefix>/lib always
but on some systems (like Fedora linux), 64bit libraries should be installed in <install_prefix>/lib64

is it ok for Java if we install into lib64 as appropriate?

PR #946, opened 2025-08-12
Update Changelog after fixing BYWEEKNO and BYSETPOS

minichma:
Update Changelog and KnownLimitations.md after #944 and #941


winterz (approved):
thanks, looks fine

PR #965, opened 2025-08-28
Fix serializing and cloning x-components

rsto:
This fixes two bugs with x-components. The `END` content line of a x-component was `END:X` instead of the name of the x-component, e.g. `END:X-FOO`. When cloning an x-component, its name wasn't set which made serialization abort.

### DIFF
--- a/src/libical/icalcomponent.c
+++ b/src/libical/icalcomponent.c
@@ -280,7 +280,7 @@ char *icalcomponent_as_ical_string_r(const icalcomponent *impl)
     }
 
     icalmemory_append_string(&buf, &buf_ptr, &buf_size, "END:");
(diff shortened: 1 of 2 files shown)


src/libical/icalcomponent.c
```diff
@@ -134,6 +134,9 @@ icalcomponent *icalcomponent_clone(const icalcomponent *old)
     icalerror_check_arg_rz((old != 0), "component");
 
     clone = icalcomponent_new_impl(old->kind);
+    if (old->x_name) {
```
winterz, line 137: seems to me that this conditional needs to be moved below the `if (clone == 0)` immediately following

rsto: Thanks, fixed.

Issue #954, opened 2025-08-21
icalrecurrencetype_rscale_is_supported()


winterz:
icalrecurrencetype_rscale_is_supported() always returns true.
I can't remember why we have this function.  should we remove it?

related question,:for the 4.0 release should we require rscale (libicu) now?
I'm ok requiring libicu for 4.0

CC: @ksmurchison ?

PR #973, opened 2025-08-31
Remove icalrecurrencetype_rscale_is_supported()

winterz:
icalrecurrencetype_rscale_is_supported always returned true because we support RSCALE=GREGORIAN even without libicu.

fixes: #954


winterz:
ignore the failing clang-tidy step.  clearly I have work to do there (need to create compile_commands.json)

minichma (approved):

Issue #984, opened 2025-09-03
ICAL_SETERROR_ISFUNC


winterz:
ICAL_SETERROR_ISFUNC is a conditional compile macro to choose between the icalerror_set_errno macro and function.

Unless someone specifically compiles with -UICAL_SETERROR_ISFUNC the default is using the function version.

consider removing this macro and making the icalerror_set_errno function 'inline'

PR #986, opened 2025-09-03
Remove the ICAL_ERRORS_ARE_FATAL conditional compile macro

winterz:
always use the icalerror_set_errno function but compile it as an inline

fixes: #984


src/libical/icalerror.c
```diff
@@ -111,8 +111,7 @@ int icalerror_get_errors_are_fatal(void)
     return icalerror_errors_are_fatal;
 }
 
-#if defined(ICAL_SETERROR_ISFUNC)
-void icalerror_set_errno(icalerrorenum x)
+inline void icalerror_set_errno(icalerrorenum x)
```
minichma, line 114: I don't know too much about how the different C compilers handle `inline` but I thought it would be required that (or at least good if) the consuming compilation units (c files) 'see' the implementation, i.e. it would be included in the header file.

But the other question would be, whether `inline` is required here at all. Should we run into an error situation, then the cost of an extra function call will probably not make any difference. On the other hand `inline` would *always* add overhead in size, so maybe its just as good to keep it as a normal function.

minichma (approved):

src/libical/icalerror.c
winterz: > I don't know too much about how the different C compilers handle `inline` but I thought it would be required that (or at least good if) the consuming compilation units (c files) 'see' the implementation, i.e. it would be included in the header file.
> 
you're right.  and I made the same mistake in 35d7dbeff497d69e5094ef7c0a66bdc0891a97e4

> But the other question would be, whether `inline` is required here at all. Should we run into an error situation, then the cost of an extra function call will probably not make any difference. On the other hand `inline` would _always_ add overhead in size, so maybe its just as good to keep it as a normal function.

I tend to agree.  I'll remove the inlining for icalerror_set_errno() in this patch.
do you think I should remove the inline pvl_data() too?

winterz: I removed the inline for  icalerror_set_errno now

minichma: > do you think I should remove the inline pvl_data() too?

Hm, as the function is exported, some compilers might complain if they see the `inline` implementation, so probably it would be good to resolve it.

In the case of `pvl_data` the benefit of `inline` might be higher compared to set_error, as it seems to be called very frequently. But I feel that the performance benefit will still be very low, probably hardly measurable. Even more so as compilers/linkers will inline anyways if they consider it beneficial.

winterz: yes, my first instinct was to not inline.  I'll remove the inline for pvl_data.

if people complain about slowness we can always revert

Issue #983, opened 2025-09-03
Split pkg-config into several components


tintou:
When one is linking against libical using pkg-config, one is actually linking to 4 libraries (`-lical -licalss -licalvcal -licalvcard`) this would be better to have one .pc file for each library

PR #988, opened 2025-09-04
pkg-config: Split the pkg-config files per library

tintou:
Allow consumers to only link the the library they need.

Closes: #983

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -992,6 +992,32 @@ if(LIBICAL_DEVMODE)
 endif()
 libical_option(LIBICAL_BUILD_EXAMPLES "Build examples." True)
 
+########### Configure pkg-config variables #########
+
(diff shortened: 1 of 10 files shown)


tintou:
Yes, the pkg-config of libicalss for instance has a `Requires: libical` so if you link against libicalss you end-up also linking against libical but you can just use libical standalone and not have any inclusion of libicalss, libicalvcal or libicalvcard as it was the case previously

mcrha:
> Yes, the pkg-config of libicalss for instance has a `Requires: libical`...

Ah, right, I did not notice it. That covers it.

Maybe some apps/users will face a problem when using libical.pc and (as it used to work) then will use a different (sub)library header (like the vcal bits when using libical.pc only), then the compiler will still find it, but the linker will have a problem. That's easy to fix on the app/user side (to use proper libical*.pc file).

Issue #848, opened 2024-12-16
Move internal functions to private API


winterz:
These functions are internal and may change at any time at our discretion.

- [ ] 

i.e. they are moved into foo_p.[c,h]

minichma:
Also related to #750

minichma:
Some candidates, but everyone TBD. Just my subjective view:

* `icaltime_adjust` (the doc comment mentions it should not be public)
* `icaltime_span_*()`: as discussed in some previous issue, not too straight forward. Especially the effect of 'isBusy' is not too obvious. Clients can easily implement the behavior as individually required.
* `icaldurationtype_from_int()` and `icaldurationtype_as_int()`: A duration cannot be directly created from seconds (and vice versa), as the date-part is treated differently (nominal) than the time part (exact). From the seconds alone its unclear what should be created, so the availability of the function could be misleading. (see #630 
* `icaldurationtype_bad_duration()` not sure
* `icalrecur_resize_by()`
* 'pvl_*`: The whole pvl.h file seems to be considered private. No public APIs use it.
* `icaltimezone_array_new`, `icaltimezone_array_append_from_vtimezone`, `icaltimezone_array_free`: internal according to comment
* `icalparser_*`: not sure, but probably internal
* `icalmemory_append_string`, `icalmemory_append_char`?
* `icalerror_stop_here`, `icalerror_crash_here`
* `icalerrno_return()`?
* `icalerrprintf`, `icalerror_warn`
* `icalerror_set_error_state`
* `icalerror_set_errno`
* `icalerror_check_*`
* `icalerror_assert`
* `icalrestriction_*`
* `sspm_*`: not sure
* `decode_quoted_printable`, `decode_base64`: not sure, but naming is not libical-like
* `icallangbind_*`: not sure 
* Everything from `astime.h`: `caldat*`, `juldat*`

Regarding `juldat*`, `caldat*`: We replaced the floating point versions with int versions some time ago and kept the previous versions for compatibility reasons. Now I realize that they aren't actually included in `ical.h`, so can't we consider them internal anyways, already today?


For all the functions we want to remove from the public API, will they still be exported? Generate a separate `ical_internal.h` from them and make clear they shouldn't be considered stable? That way we would still allow users to use them if they are ready to adjust if the API changes.

winterz:
> For all the functions we want to remove from the public API, will they still be exported? 

yes, when necessary

[...]

PR #989, opened 2025-09-04
Move some icaltime functions to private API

winterz:
Move some icaltime functions to private API

Moved to icaltime_p.[c,h]:
 * The icaltime_t type
 * The icaltime_span type
 * icaltime_span_new
 * icaltime_span_overlaps

GH-848


minichma:
@winterz I will try to have a look later today, tomorrow at the latest.

src/libical/icaltime.h
```diff
@@ -72,9 +72,10 @@
 
 #include "libical_ical_export.h"
 
+#include "icaltime_p.h"
```
minichma, line 75: Do I understand correctly, that the `*_p.h` files are installed besides `ical.h`? By including them from the central `ical.h` people have the private functions available by just including the public `ical.h`. It will not be obvious to them that certain functions are intended only for internal use while most others are not. Wouldn't it be cleaner *not* to include the *_p.h files from the central `ical.h` (i.e. not include `icaltime_p.h` from `icaltime.h`)? Instead the private *_p.h files could perfectly include the public `ical.h` file. Should people intend to use private functions, they'd have to _explicitly_ include the private header files and it would be clear that those are subject to change.
It would also be less likely to happen that we unintentionally use private types within the public declarations.

minichma (commented):
Here are some first thoughts on the overall structure of the private headers. Please let me know what you think!

PR #993, opened 2025-09-08
Move some icalduration functions to private API

winterz:
Moved to private API:
* icaldurationtype_as_int
* icaldurationtype_from_int


winterz:
should be ready for review now.  
@minichma does this feel better than what I did for the private icaltime functions?

minichma:
@winterz sorry to say that I won't be able to have a closer look before the weekend

minichma (approved):
@winterz I feel this approach is much cleaner. Reflecting to [@mcrha's comments](https://github.com/libical/libical/issues/848#issuecomment-3269127469) - this approach allows moving functions back to the public API at any time should we find out that we hid too much.

Issue #713, opened 2024-06-15
Clean '#if 0'


winterz:
remove the old '#if 0' conditional compiles

PR #997, opened 2025-09-08
cleanup: Remove #if 0 in the code

mcrha:
The code is not used, thus drop it.

Closes https://github.com/libical/libical/issues/713


winterz:
> > please also remove the `#if 0` in `src/libicalss/icalssyacc.c`
> 
> That one seems to be a generated file and it is used for translations, as the comment there says. I can drop it, libical does not have gettext translations (if I'm not mistaken), then it does not matter, just as soon as the file is regenerated, it'll be back.

ok.

Issue #844, opened 2024-12-13
Missing libical-glib API files for vcard library


winterz:
We need to add the libicalvcard methods for the glib introspection

mcrha:
I can look on this, but I'd like to coordinate how to do it first, to pick the right approach.

* A quick attempt reveals some obstacles in a natural way, which would be to create a `src/libcalvcard-glib` directory and place the API definition files and the build files into it. I do not want to duplicate the generator, but it has some in-code references to "ical", which makes it less generic than I thought. Either the generator can be changed to extract these things into app parameters or the generator code can be copied (I do not like any duplication)
* alternatively, the libicalvcard API can be added into the `libical-glib` itself, which can help on couple places, but won't be the cleanest solution; on the other hand, if you consider `libical-glib` as the glib bindings for "everything" what the libical can offer, then it's all fine. The other bindings do not split too, though they seem to do only the iCalendar part at the moment. (Honestly, vCard inside iCalendar library feels kinda odd.)
* while the `libicalvcard` uses namescape `vcard`. I'd rather make it `i_cal_vcard_...`/`LibIcalVcard` for the glib bindings, to ensure it does not clash with other vCard implementation(s).

winterz:
> Looking how the header files from the other sub-libraries are added into the `$PREFIX/include/libical/`, I suppose having everything in the `src/libical-glib/` is fine.

I agree. 

> while the libicalvcard uses namescape vcard. I'd rather make it i_cal_vcard_.../LibIcalVcard for the glib bindings, to ensure it does not clash with other vCard implementation(s).

whatever you think is best works for me.

tintou:
I don't think that there is any other vcard library in the GLib space so you can use the v_card namespace as the C namespace is already using it anyway

tintou:
@mcrha You need to wrap these types as the type system has no information on how to allocate/copy/free these structures, also we are only introspecting one namespace so vcard_* won't be part of it.

tintou:
Maybe PyGObject allows you to use it with `_30` or `@30`, you can see with `print(list(ICalGLib.VcardPropertyVersion))` I think

mcrha:
I've a patch ready in https://github.com/mcrha/libical/tree/wip/libicalvcard-glib , only issing thing is about those `VERSION_nn`, which break the usage (it builds fine, but when one tries to use it it fails; see above).

mcrha:
[...]

PR #999, opened 2025-09-09
libical-glib: Add bindings for libicalvcard

mcrha:
Closes https://github.com/libical/libical/issues/844


tintou:
Can we have two libraries here? ical-glib and ical-vcard-glib

tintou:
Okay, I'll try to give it a look before the release then :slightly_smiling_face:

winterz:
@mcrha or @tintou feel free to merge this one whenever you think its ready to go.

my only thought is: I'd rather not have 2 libraries ( ical-glib and ical-vcard-glib) if that can be avoided.

PR #1029, opened 2025-09-18
builtin_timezones concurrency fixes

ksuther:
I might be hunting in the wrong place, but I think there's a race when calling `icaltimezone_get_builtin_timezone` on multiple threads and `builtin_timezones` is loading for the first time. I didn't have any luck triggering it with clang's thread sanitizer enabled though.

We're seeing random crashes when calling `icaltimezone_get_builtin_timezone` concurrently:
```
Exception Type: EXC_BAD_ACCESS (SIGSEGV)
Exception Codes: SEGV_NOOP
Crashed Thread: 20

Application Specific Information:
Africa/Luanda > America/Halifax > comp >
Attempted to dereference null pointer.

Thread 20 Crashed:
0   libsystem_platform.dylib        0x53ec5d084         _platform_strcmp$VARIANT$Base
1   libical                         0x104bcf4a8         icaltimezone_get_builtin_timezone (icaltimezone.c:1379)
...remainder of stack trace...
```

Line 1379 in our fork is this line: https://github.com/libical/libical/blob/ae628a34a675bce5ddae3146cafee36714415578/src/libical/icaltimezone.c#L1362

What I think is happening:
- Thread 1 calls `icaltimezone_get_builtin_timezone` and `builtin_timezones` is NULL, so it calls `icaltimezone_init_builtin_timezones,` then `icaltimezone_parse_zone_tab`
- Thread 1 holds `icaltimezone_builtin_lock`
- Thread 1 allocs `builtin_timezones` inside `icaltimezone_parse_zone_tab`, but hasn't yet added any timezones to the array
- Thread 2 calls `icaltimezone_get_builtin_timezone` and `builtin_timezones` is not NULL, but doesn't hold any valid time zones yet
- Thread 2 tries to access `builtin_timezones`, which hasn't finished loading on Thread 1, and crashes

This fixes the problem by only setting `builtin_timezones` on Thread 1 when it has finished loading. This means that Thread 2 would see `builtin_timezones` is NULL, then call `icaltimezone_init_builtin_timezones`, hit the lock, wait for Thread 1 to finish loading, then continue because the second check if `builtin_timezones` is NULL would fail.

I also noticed that `icaltimezone_release_zone_tab` wasn't using the lock, so I added that.


winterz:
and one from `test/icaltm_test`

```
#0  0x00007f2d1af0126b in icaltimezone_get_builtin_timezone (location=0x4013a0 "America/New_York") at /home/winterz/projects/libical/libical/src/libical/icaltimezone.c:1359
1359        for (lower = 0; lower < builtin_timezones->num_elements; lower++) {
#0  0x00007f2d1af0126b in icaltimezone_get_builtin_timezone (location=0x4013a0 "America/New_York") at /home/winterz/projects/libical/libical/src/libical/icaltimezone.c:1359
```

CMendia:
@ksuther It only fails on Windows because builtin timezone data is only used on Windows (search CMakeLists.txt for "builtin"). You can add `#define USE_BUILTIN_TZDATA` to icaltimezone.c to get the bug.

ksuther:
Thanks for pointing that out. The issue was that in the error cases `builtin_timezones` wasn't being set. The tests are now passing.

I also noticed that `icaltimezone_free_builtin_timezones` is missing a lock. Is there a reason that function exists separately from `icaltimezone_release_zone_tab`? It seems like they both do the same thing, except `icaltimezone_free_builtin_timezones` does it in a more verbose way.

winterz (approved):
looks good to me

winterz:
> I also noticed that `icaltimezone_free_builtin_timezones` is missing a lock. Is there a reason that function exists separately from `icaltimezone_release_zone_tab`? It seems like they both do the same thing, except `icaltimezone_free_builtin_timezones` does it in a more verbose way.

I have no memory of how those functions became what they are today.
and it seems to me that a lock is good to have and should not be avoided with a  call to `icaltimezone_free_builtin_timezones`

but nobody is calling `icaltimezone_release_zone_tab`

@mcrha do GNOME applications use `icaltimezone_release_zone_tab`?

... feels like we can replace calling `icaltimezone_free_builtin_timezones` with ``icaltimezone_release_zone_tab` everywhere and remove the `icaltimezone_free_builtin_timezones` definition. but I haven't tested that at all.

thoughts?

mcrha:
I tried `git grep` in the evolution-data-server and evolution itself and none of them call any of these two functions (neither the libical version nor the libical-glib version). These want the timezone to live ad infinity, because they are (re-)used all around the places. Easier than to face a use-after-free on app quit/close "accidentally", I think.

ksuther:
I also added a lock to `icaltimezone_free_builtin_timezones`. Deleting `icaltimezone_release_zone_tab` seems reasonable to me since there's an obvious equivalent if you really want it. I can't imagine many people are intentionally clearing the time zone data anyway.

Should I remove it in this PR or leave it for something separate?

Issue #476, opened 2021-01-29
ICAL_ALLOW_EMPTY_PROPERTIES=true allows invalid events


ksmurchison:
Enabling empty properties causes an event such as the following to be accepted by libical

BEGIN:VCALENDAR
VERSION:2.0
PRODID:-//FooBar/1.0/EN
BEGIN:VEVENT
DTEND:20210128T203544
DTSTAMP:20210128T193544Z
DTSTART;TZID=America/New_York:
SUMMARY:hiya
UID:039D51F6-61A0-11EB-800B-79434D0E0A52
END:VEVENT
END:VCALENDAR

I *think* the intention of ICAL_ALLOW_EMPTY_PROPERTIES was for text values such as DESCRIPTION or TZID.  I don't think it should apply to properties such as DTSTART.  The group of properties that fall in this category is up for debate.

winterz:
In 3.1 (or 4.0 .. i.e. current master) how about we replace allow_empty with strictness levels.
strictness_level=0 (most strict, what we have now by default)
strictness_level=1 (allow empty properties DESCRIPTION, TZID, )
strictness_level=2 (allow empty UUID and create one for you)

... something like this.   that gives us flexibility for future easing

there would be API for getting/setting the strictness level.

spit-balling

ksmurchison:
I was also thinking about making this a runtime option, regardless of what we [don't] allow

ksuther:
Is there anything in the spec that says empty text properties are illegal? The reason I originally added the flag in the first place was that we were encountering ics files with blank `SUMMARY`, `LOCATION`, and `DESCRIPTION` properties. allowing empty `DTSTART` properties is an unintended side effect.

The strictness level makes sense and would allow us to have a matrix of options similar to what @ksmurchison did for the RRULE leniency. Having it also do stuff like empty UUID checking would worry me a little because you might want those to be independent flags rather than additive as separate levels.

ksmurchison:
RFC 5545 defines text as:

text       = *(TSAFE-CHAR / ":" / DQUOTE / ESCAPED-CHAR)

So it's legal to be empty AFAICT. Structured types shouldn't be allowed to be empty IMHO unless explicitly allowed by the spec

winterz:
more spit-balling many years later...

would it be enough to add an `icalproperty_is_structured` function (we already have `vcardproperty_is_structured`)
and then error if value is empty and icalproperty_is_structured?

[...]

PR #1040, opened 2025-09-25
Add new publicly visible function icalproperty_is_structured()

winterz:
public to follow vcardproperty_is_structured()


ksuther (approved):
Are there any examples of scenarios where this would be useful other than #476? Just asking in case it gives someone else usage ideas.

The code looks good to me otherwise.

PR #1045, opened 2025-09-28
icalduration: from/as_int to from/as_seconds

### LINKED ISSUE #1044: Rename  icaldurationtype_as_int and  icaldurationtype_from_int
@CMendia writes 

> These functions are really misnamed: icalduration_as_int should be icalduration_as_seconds and icalduration_from_int should be icalduration_from_seconds. With their current names it they seems like they encode a duration as an integer. If we do decide that we should be able to encode durations as integers, we should also consider whether other data structures need to be similarly encoded, for consistency if nothing else.

rather making these private, perhaps we improve the naming and the documentation and explain what not to use them for.

CMendia:
Change icalduration_from_int and icalduration_as_int to icalduration_from_seconds and icalduration_as_seconds, respectively. Fixes #1044.

This is essentially a global search and replace of these two names, except the code in [src/libical/icalvalue.c](https://github.com/libical/libical/compare/master...CMendia:seconds?expand=1#diff-1dd2e567bcf8e9ee5fbfc53404fe5907d3bab50954eaafcc0c30cbdbede810fd), which is new. The new code fixes comparison of nominal and accurate durations.

Because of #202, I have ignored most of the Perl code that still needs to be changed.

### DIFF
--- a/src/libical/icalvalue.c
+++ b/src/libical/icalvalue.c
@@ -1398,15 +1398,32 @@ icalparameter_xliccomparetype icalvalue_compare(const icalvalue *a, const icalva
     }
 
     case ICAL_DURATION_VALUE: {
-        int dur_a = icaldurationtype_as_int(a->data.v_duration);
-        int dur_b = icaldurationtype_as_int(b->data.v_duration);
+        struct icaldurationtype dur_a = a->data.v_duration,
+                                dur_b = b->data.v_duration;
+        int a_days = (int)(7 * dur_a.weeks + dur_a.days) * (dur_a.is_neg == 1 ? -1 : 1),
+            b_days = (int)(7 * dur_b.weeks + dur_b.days) * (dur_a.is_neg == 1 ? -1 : 1);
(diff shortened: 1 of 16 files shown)


src/libical/icalvalue.c
```diff
@@ -1398,15 +1398,34 @@ icalparameter_xliccomparetype icalvalue_compare(const icalvalue *a, const icalva
...
+                return ICAL_XLICCOMPARETYPE_GREATER;
+            } else if (a_days < b_days) {
+                return ICAL_XLICCOMPARETYPE_LESS;
+            } else {
+                return ICAL_XLICCOMPARETYPE_EQUAL;
+            }
+        } else if (a_days == b_days) {
+            if (a_secs > b_secs) {
+                return ICAL_XLICCOMPARETYPE_GREATER;
+            } else if (a_secs < b_secs) {
```
winterz, line 1421: cppcheck says
"icalvalue.c:1421,style,knownConditionTrueFalse,Condition 'a_secs<b_secs' is always true"

CMendia: Ah, true. Thanks.

PR #1058, opened 2025-10-02
expand_year_days: Verify month is in range

tsdgeos:
oss-fuzz creates some files where the month is 4000 and then ends up in icaltime_day_of_year (without icu) that ends up doing an out of bounds array access


winterz:
thanks @tsdgeos 
btw: where did you encounter this fuzz issue?  I don't see it at the libical project on oss-fuzz.com

Issue #462, opened 2020-12-14
Complete quoting for parameters of some CMake commands


elfring:
Some parameters (like “[`${CMAKE_CURRENT_BINARY_DIR}/config.h`](https://github.com/libical/libical/blob/555a7d3e620c48acd625218ada565ea4b5eee921/CMakeLists.txt#L321 "Update candidate")”) are passed to CMake commands in your build scripts without enclosing them by quotation marks. I see that these places will result in build difficulties if the contents of the used variables will contain special characters like semicolons.

I would recommend to apply [advices from the software documentation](https://cmake.org/cmake/help/v3.19/manual/cmake-language.7.html#lists "Description for CMake lists").

winterz:
good point

winterz:
no doubt.  I wonder if there's a linting tool to help find where the double-quotes are missing.
otherwise, I'd just be blindly adding them all over the place

winterz:
for #1071 I did a lot of testing with whitespace in the build and installation dirs.
next I will test with whitespace in the source dir.

PR #1071, opened 2025-10-06
Buildsystem - handle whitespace in paths

winterz:
fixes: #462

### DIFF
--- a/.gersemirc
+++ b/.gersemirc
@@ -2,3 +2,4 @@ definitions: [./cmake,./src/libical/,./src/test,./docs]
 indent: 2
 line_length: 120
 list_expansion: favour-expansion
+warn_about_unknown_commands: false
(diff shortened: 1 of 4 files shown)


winterz:
are you saying that  `set(test_args "-f \"${PROJECT_BINARY_DIR}/src/test/${reference_data}\"")`
would be better written as `set(test_args "-f ;${PROJECT_BINARY_DIR}/src/test/${reference_data}")`
or?

winterz:
I tried various things and I couldn't make any improvements on this patch.

`add_test` creates a  `run_test.cmake` script which calls `separate_arguments` which needs a whitespace separated string of arguments

if you have a better idea please give me a specific solution.

elfring:
:thought_balloon: Better quoting might not only be helpful for the handling of development concerns with white-space characters.

PR #1072, opened 2025-10-06
xmllint all xml files and add pre-commit to enforce that

winterz:


mcrha:
I do not know how these precommit-s work, it's done only on the server? Where do you set the xmllint dependency, apart of the reference to the config file? Or I guess it's done transparently, I do not need to care?

winterz:
> I do not know how these precommit-s work, it's done only on the server? Where do you set the xmllint dependency, apart of the reference to the config file? Or I guess it's done transparently, I do not need to care?

yes. github will take care of fixing for you, when possible.
if you want to fix beforehand (during the `git commit`) you can install pre-commit locally:

- install pre-commit `sudo dnf install pre-commit`
- cd /path/to/libical-source-code
- run `pre-commit install`

btw: since the pre-commit.ci system doesn't have xmllint I had to switch to another tool (lxml)

mcrha:
Thanks for the pointers.

PR #1077, opened 2025-10-08
vcardcomponent.c: don't add/remove urn:uuid: prefix on UID

ksmurchison:
fiddling with UID has proven problematic; if its text, leave it as text


rsto (approved):
Definitely better than the current "urn:uuid" rewrite. That being said: we are in the process of testing vCard version 3 and 4 conversion more thoroughly, so while that PR looks better than the currently merged code, it might not be the end of the story.
I agree that adding a test would be good, especially for detecting later regressions.

PR #1106, opened 2025-10-14
vcardcomponent_get_version() - return VCARD_VERSION_NONE on failure

winterz:
Instead of returning 0 (not a legit vcardproperty_version) return VCARD_VERSION_NONE if a null card pointer is provided.


rsto (approved):
Agreed, that's better.

PR #1116, opened 2025-10-18
enforce generated code to follow our coding standards

winterz:
readability-braces-around-statements

not all fixed since there is some generated code in libical-glib that I don't know how to fix yet.


winterz:
but I will merge this PR and we can work on the final touches in followups.

winterz:
inside your build directory run 
```
clang-tidy --checks=readability-braces-around-statements `find . -name "*.c" -o -name "*.h"`
```

Issue #345, opened 2018-05-09
icalrestrction.c: VEVENT -> ICAL_RESTRICTION_ZEROORONE or ICAL_RESTRICTION_ONE


dilyanpalauzov:
My `libical/src/libical/icalrestrction.c` contains:
```
    {ICAL_METHOD_NONE, ICAL_VCALENDAR_COMPONENT, ICAL_UID_PROPERTY, ICAL_RESTRICTION_ZEROORONE, NULL},
    {ICAL_METHOD_NONE, ICAL_VEVENT_COMPONENT, ICAL_UID_PROPERTY, ICAL_RESTRICTION_ZEROORONE, NULL},
    {ICAL_METHOD_NONE, ICAL_VTODO_COMPONENT, ICAL_UID_PROPERTY, ICAL_RESTRICTION_ZEROORONE, NULL},
    {ICAL_METHOD_NONE, ICAL_VJOURNAL_COMPONENT, ICAL_UID_PROPERTY, ICAL_RESTRICTION_ZEROORONE, NULL},
    {ICAL_METHOD_NONE, ICAL_VFREEBUSY_COMPONENT, ICAL_UID_PROPERTY, ICAL_RESTRICTION_ONE, NULL},
    {ICAL_METHOD_NONE, ICAL_VTIMEZONE_COMPONENT, ICAL_UID_PROPERTY, ICAL_RESTRICTION_ZERO, NULL},
    {ICAL_METHOD_NONE, ICAL_XSTANDARD_COMPONENT, ICAL_UID_PROPERTY, ICAL_RESTRICTION_ZERO, NULL},
    {ICAL_METHOD_NONE, ICAL_XDAYLIGHT_COMPONENT, ICAL_UID_PROPERTY, ICAL_RESTRICTION_ZERO, NULL},
    {ICAL_METHOD_NONE, ICAL_XAUDIOALARM_COMPONENT, ICAL_UID_PROPERTY, ICAL_RESTRICTION_ZERO, NULL},
    {ICAL_METHOD_NONE, ICAL_XDISPLAYALARM_COMPONENT, ICAL_UID_PROPERTY, ICAL_RESTRICTION_ZERO, NULL},
    {ICAL_METHOD_NONE, ICAL_XEMAILALARM_COMPONENT, ICAL_UID_PROPERTY, ICAL_RESTRICTION_ZERO, NULL},
[...]

PR #1146, opened 2025-11-14
WIP - Require UI and DTSTAMP for VEVENT, VTODO and VJOURNAL

winterz:
To conform to RFC5545

fixes: #345


winterz:
I marked this as a WIP since I'm looking for a discussion and opinions.

winterz:
then I think we're back to my original idea:  document were we're not 100% compliant and let the applications deal with the fall-out.

ksuther:
That makes sense to me. Seems like a proper application of Postel's Law (be conservative in what you send, be liberal in what you accept, https://en.wikipedia.org/wiki/Robustness_principle)

PR #1153, opened 2025-11-25
Use the CMake variable CMAKE_C_BYTE_ORDER to detect ENDIANness

winterz:
part of the fix to simplify the byte-swapping code in icaltimezone_p.c

GH:711


winterz:
macos-13 brownout is in-progress
look again tomorrow.

PR #1178, opened 2026-01-02
Fix an oss-fuzzer issue by limiting the number of search iterations

winterz:
In icalrecur_iterator_next(), limit the number of search iterations (is 10 million enough?) before giving up finding the next occurence.


ksuther (approved):
10,000,000 certainly seems like enough to me (27,397 years of daily recurring events).

rsto:
Thanks for this @winterz. I agree that the hard-coded limit looks sane, if need be, we can still make it a build or runtime option later.

Issue #1173, opened 2025-12-22
Builtin time zone offsets for historic dates are wrong


smartmic:
Hey there,
I was puzzled by a difference between a program I wrote for both Windows and Linux. 
As test data, I use some ics files with quite historic event dates.
On Windows, I use the builtin time zone data, on Linux the system zone info.
Now the difference is in time zone offsets. 
The issue is that the builtin time zone data base does not include previous conventions.
For example, the date `DTSTART;TZID=America/New_York:19980312T083000` is in EST (UTC-5) as of 1998, but libical "builtin" interprets it as EDT (UTC-4) due to the VTIMEZONE rule in *zoneinfo/America/New_York.ics*.
I do not say libical should include historic offsets (though it would be possible via different VTIMEZONE), but formally, shouldn't the DTSTART of the VTIMEZONE corrected?
According [Wikipedia](https://en.wikipedia.org/wiki/Daylight_saving_time_in_the_United_States), the relevant rule for my test data changed in 2007 whereas the VTIMEZONE ics says `DTSTART:19700308T020000`:

> By the [Energy Policy Act of 2005](https://en.wikipedia.org/wiki/Energy_Policy_Act_of_2005), daylight saving time (DST) was extended in the United States beginning in 2007.[[22]](https://en.wikipedia.org/wiki/Daylight_saving_time_in_the_United_States#cite_note-22) From that year, DST begins on the second Sunday of March and ends on the first Sunday of November.[[23]](https://en.wikipedia.org/wiki/Daylight_saving_time_in_the_United_States#cite_note-23)

If advised, I could prepare a patch at least for US where the DTSTART of the VTIMEZONE is corrected for this. 
However, I have spotted problems in other time zones as well. 
That leads me to the question: where you are sourcing the tz2024a data from? Did I got something wrong?

winterz:
yes please make a pull request. 

@ksmurchison @ksuther any thoughts?

PR #1190, opened 2026-01-06
Update zoneinfo to tzdata2025c including historical data

smartmic:
This is in reference to #1173, the new zoneinfo, which includes historical offsets. It was generated by `vzic --pure` and is based on tzdata2025c, which can be found at https://ftp.iana.org/tz/releases/tzdata2025c.tar.gz.

My fork branched from libical 3.0, so I targeted this branch. However, it should also be applicable to master and other branches.

As written in the issue (and vzic documentation), MS Outlook compatibility using `vzic --pure` might be impacted. Please judge based on your experience and test coverage.

### DIFF
--- a/ReleaseNotes.txt
+++ b/ReleaseNotes.txt
@@ -3,6 +3,7 @@ Release Highlights
 
 Version 3.0.21 (UNRELEASED):
 ----------------------------
+ * Built-in timezones updated to tzdata2025c including historical offsets
  * Fix some scenarios with RRULEs and EXRULE (#754)
 
 Version 3.0.20 (10 March 2025):


winterz:
> Otherwise, I have nothing against the suggestion to use the 4.0 branch as a target. I would then use my 3.0 fork for my application development purposes as long as I have not migrated to 4.0 and value the historical offsets.

ok, I'd feel better reverting in 3.0 .  I'm not even sure I want another 3.0.x release..  I'd rather put our limited time and resources into the 4.0 release.  but, I think for 4.0 we can just go ahead and push this out and if it breaks in the wild then we'll know about it and can deal with the consequences.  and if you want to test in the meantime that would also be great.

PR #1219, opened 2026-01-28
icalparser.c - limit line length in parser_get_next to 10000 chars

winterz:
To fix an oss-fuzz issue for 60 cpu timeout we reduce the maximum number of chars we can search in parser_get_next() from 100000 to 10000 chars.

https://oss-fuzz.com/testcase-detail/6015910128451584


winterz:
in the meantime, somehow/someway the fuzz issue is fixed.
closing

PR #1222, opened 2026-02-01
Fix vCard parser looping endlessly for dangling backslash

rsto:
Fixes a bug in the vCard parser detected in https://github.com/libical/libical/pull/1221


winterz:
the fix was committed a week ago in e7b479b983ae6de2e9f0e38b23ec1fec1cfda0e7

the other test code is part of another patch that's waiting to merge

Issue #1223, opened 2026-02-01
BYWEEKNO RRULE returns different results for different DTSTART where it shouldn't


minichma:
The following test case is part of [`icalrecur_test.txt`](https://github.com/libical/libical/blob/e7b479b983ae6de2e9f0e38b23ec1fec1cfda0e7/src/test/icalrecur_test.txt#L660-L663) (seemingly related to #1143):

```ics
# github issue1143
RRULE:FREQ=YEARLY;BYWEEKNO=53;COUNT=1
DTSTART:20230101T000000Z
INSTANCES:20270103T000000Z
```

The DTSTART value itself is not part of the expected instances, which is an undefined case according to the RFC (but doesn't mean we can't write test cases for it). However, when specifying the given instance as DTSTART, the test fails, which is unexpected:

```ics
RRULE:FREQ=YEARLY;BYWEEKNO=53;COUNT=1
DTSTART:20270103T000000Z
Expected INSTANCES:20270103T000000Z
Actual   INSTANCES:20330102T000000Z
```

This ticket doesn't say anything about the above tests being right or wrong. It just addresses the fact, that the two test cases return different results where they seemingly shouldn't.

PR #1224, opened 2026-02-01
icalrecur_test: Extend test execution to attempt every test recurrence instance as DTSTART

minichma:
It should be possible for any RRULE, to specify any recurrence instance as DTSTART and get that exact DTSTART as well as all following ones as recurrence instances.

This PR extends `icalrecur_test` to run the individual test cases from the `icalrecur_test.txt` file with each of the expected recurrence instances as DTSTART to cover the above expectation.

The branch finds two (related) test cases that violate that assumption, which is documented in #1223.

This PR should be considered for merging after #1223 was fixed.

### DIFF
--- a/src/test/icalrecur_test.c
+++ b/src/test/icalrecur_test.c
@@ -171,7 +171,6 @@ static int run_testcase(struct recur *r, bool verbose, bool forward, int proceed
     const char *sep = "";
     char actual_instances[2048];
     int actual_instances_len = 0;
-    struct icaltimetype start = icaltime_null_time();
     int test_error = 0;
 
     if (verbose) {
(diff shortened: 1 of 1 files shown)


minichma:
Hm, the 'regular' recurrence tests succeed now, but the RSCALE tests still fail. `SKIP=FORWARD/BACKWARD` doesn't work as I'd have expected. I'd assume that also in case of RSCALE/SKIP the assumption should hold that it should be possible to use returned recurrence instances of an RRULE also as DTSTART for that RRULE. Right? But this doesn't seem to be the case.

CMendia:
I don't know that we should expect a rule to produce the same results when its first instance is set as the DTSTART if that instance comes about because of a SKIP=FORWARD property. The date to which we skip is outside the rule, so starting from that date (instead of an earlier date) should yield different results.

winterz:
src/test/icalrecur_test.c:174:25: warning: Value stored to 'start' during its initialization is never read [deadcode.DeadStores]

please fix that little nit and then you can merge

PR #1229, opened 2026-02-08
Fixes #1213 (issue with BYWEEKNO 53 when DTSTART is an expected instance)

### LINKED ISSUE #1223: BYWEEKNO RRULE returns different results for different DTSTART where it shouldn't

CMendia:
Fixes #1223 by expanding a mask for the previous year if our start date is in the first week of the previous year.

### DIFF
--- a/src/libical/icalrecur.c
+++ b/src/libical/icalrecur.c
@@ -3827,14 +3827,14 @@ static bool __iterator_set_start(icalrecur_iterator *impl, icaltimetype start)
         /* if we are in the last week of the previous year,
          * expand year days for the previous year
          */
-        if(has_by_data(impl, ICAL_BY_WEEK_NO) &&
-                get_week_number(impl, start) == 53 &&
-                start.month == 1) {
+        if (has_by_data(impl, ICAL_BY_WEEK_NO) &&
+            get_week_number(impl, start) == 53 &&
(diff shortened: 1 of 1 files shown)


CMendia:
(1213 in the title should be 1223.)

PR #1258, opened 2026-02-25
Bring back vcardfuzzer test 485932113

winterz:


winterz:
yay.  this one works now.
I'd like to think we improved/fixed something in the code that caused windows to fail.

however, I won't merge this since clearly has a long run time and nobody will want to run test it locally and it will slow down the CI responses as well.

=> closing (but happy the CI is green)

PR #1268, opened 2026-03-09
icalcomponent.c: do not enforce const pointer on foreach_recurrence

rsto:
This reverts 846744148, which had redeclared the icalcomponent argument of the `foreach_recurrence` callback to be a const pointer. While declaring arguments as const in general is a good choice, it isn't in this particular case:

Calling any of the `icalcomponent_get_*` functions fails due to their icalcomponent argument being a non-const pointer. Similarly, iterating the component is not possible. Redeclaring all these functions to take const arguments requires more work. For example, the `icalcomponent_get_timezone` function currently requires the icalcomponent argument to be mutable, as it otherwise could not cache timezones.

### DIFF
--- a/src/libical/icalrecur.c
+++ b/src/libical/icalrecur.c
@@ -3867,8 +3867,8 @@ static bool __iterator_set_start(icalrecur_iterator *impl, icaltimetype start)
 
         if (has_by_data(impl, ICAL_BY_WEEK_NO)) {
             int start_weekno = get_week_number(impl, start);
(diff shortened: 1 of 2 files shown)


rsto:
> I mean: make a new patch with a regression test. which should fail until the revert lands

I can't provide a test that only fails at runtime for this. The closest one can get is to write a test that attempts to call `icalcomponent_get_uid` or the like with the const argument. But that will already fail during build.

PR #1269, opened 2026-03-10
Support components and properties with iana-token name

rsto:
This adds support to parse and create iCalendar components and properties with iana-token name. Before, these got discarded during the parse. Now they get parsed if the `ical_unknown_token_handling` is set to `ICAL_ASSUME_IANA_TOKEN`.

This PR is still a draft until I've finished some more integration testing. I figured it might be worthwhile posting this now to show where this is going.


winterz:
so far looks good to me.

PR #1273, opened 2026-03-12
Handle extension parameter and property names case-insensitively

rsto:
RFC 5545, Section 3.1 defines parameter and property names to be case-insensitive, so don't require extension parameters and properties to start with capital "X-" and compare names case-insensitively.

Also fixes `icalproperty_remove_parameter_by_kind` and `icalproperty_remove_parameter_by_name` for which the documentation claimed to remove *all* parameters, but the implementation only removed the first matching parameter.


winterz:
> Oh, you mean to document the change for parsing case-sensitive `x-` names?

yes, mention that we now follow RFC 5545, Section 3.1 for X- names too (i.e they don't need to be uppercase any longer)

PR #1291, opened 2026-04-12
RRULE parsing: Be more strict against overflows and empty values

minichma:
So far the RRULE parser didn't check for overflows when parsing `BYDAY` integers. This was partly intentional to support rule parts like `BYDAY=255SU`,  (which used to be sent by Novell/GroupWise instead of `BYDAY=-1SU`). As these clearly violates the RFC and don't seem to too relevant these days, this PR updates the parser to consider such overflows as errors.

It also is more strict against empty strings when parsing `BY` rule parts, which previously interpreted empty strings as 0.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -54,6 +54,8 @@ for details about API changes since libical 3.x.
   are parsed as X kind.
 - The `icalproperty_remove_parameter_by_(kind|name)` functions now remove all matching parameters.
   Before, they only removed the first matching parameter contrary to their documentation.
+- `icalrecurrencetype_new_from_string()` is more strict against overflows and empty rule parts.
+  E.g. it now rejects RRULE parts like `BYDAY=255SU` (which used to be interpreted as `BYDAY=-1SU`).
 
 ### Deprecated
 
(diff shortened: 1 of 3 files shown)


minichma:
@winterz I stumbled across these seeming shortcomings of the RRULE parser when working on some tests and created a fix, just to note that the previous overflow handling was intentional (at least partly). See

https://github.com/libical/libical/blob/2d5b2f393d00d7ea663fb290f2bf9f23c055c04e/src/libical/icalrecur.c#L654-L655

However, I feel these days it would be better to be standards-compliant than supporting legacy systems. At least in the upcoming v4. What do you think?

winterz:
we have a memory leak now in icalrecur_add_bydayrules.
you need to add `icalmemory_free_buffer(vals_copy);` before the new `return -1` lines


also: cppcheck is reporting 
icalrecur.c:964,style,knownConditionTrueFalse,Condition 'parser.rt->interval!=tmp' is always false

please add the line
/* cppcheck-suppress knownConditionTrueFalse */ immediately above line 964

winterz (approved):
all good now.
you can merge.

minichma:
thank you!

PR #1292, opened 2026-04-12
CMakePresets: Add `configurePresets` `dev-windows`

minichma:
Add CMake presets for development on windows using the MSVC toolchain. It is based on the settings used in the `build-windows` GitHub Workflow.


winterz (approved):
sure. why not.

PR #1294, opened 2026-04-12
Don't try to build libical-glib-build-check on Windows with MSVC

winterz:
MSVC doesn't have -Wswitch-enum and -Wswitch options.


winterz:
@mcrha unless you want to put work into figuring out how to build on Windows with MSVC I think this is good enough.

mcrha:
Works for me :)

PR #1299, opened 2026-04-13
CMake - no longer use our hand-built FindGLIb

winterz:


tintou:
Don't forget https://github.com/libical/libical/blob/848ddeacda70cff8f4fc2674ac7bf4cefc16a790/vzic/CMakeLists-integrated.txt#L57

winterz:
@tintou can you take a look?

I had to call pkg_search_module twice on Mac, for unknown reasons.  maybe due to homebrew

PR #1300, opened 2026-04-13
CHANGELOG.md - Discuss the libicalvcard Technical Preview

### LINKED ISSUE #1262: Remove libicalvcard from v4 release
We (Fastmail, notably @ksmurchison and me) contributed libicalvcard a while ago to the main branch to allow processing vCard data using the same idioms like libical already provides for iCalendar. However, we now suggest to not ship it in the upcoming libical version 4 release.

While the library is mostly complete, it not necessarily is robust enough for a stable release. We successfully used it in our CardDAV implementation since over 2 years but only very recently started using it also for JMAP for Contacts. It's likely that we'll have to tweak the code to accommodate for edge cases that we'll encounter in the next months. Shipping it as-is in libical version 4 would require us to commit to only introduce non-breaking changes. To us, it doesn't look like the right time for this.

It currently also looks like no one besides Fastmail already uses libicalvcard, so I guess removing it from the v4 milestone won't inconvenience anyone? There is an open issue to deprecate libicalvcal in favor of libicalvcard (https://github.com/libical/libical/issues/763), but does any of the projects currently using libicalvcal actually plan to start using libicalvcard?

winterz:
fixes: #1262

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,12 +12,22 @@ This is a major release and is **not** source or binary compatible with version 
 Please see [Version 4 Migration Guide](MigrationGuide_to_4.md)
 for details about API changes since libical 3.x.
 
+Included in this release is a **Technical Preview (TP)** of our new library libicalvcard
+for handling VCARD formatted data.  We encourage users to try-out this new library
+and provide feedback to our issue tracker at <https://github.com/libical/libical/issues>.
+Note that the libicalvcard API is not finalized and no source or binary compatibility
+is guaranteed until sometime later in 4.x release cycle. Additionally, there are currently
+no C++ bindings available for libicalvcard (volunteers wanted).
+
+Please **do not** use for libicalvcard for production or other non-experimental purposes
+until the Technical Preview phase is completed.
+
 ### Added
 
 - REUSE compliant licensing
 - Supports multi-valued parameters (DELEGATED-FROM, DELEGATED-TO, MEMBER, DISPLAY, FEATURE, etc)
 - Adds java bindings
-- Brand new icalvcard library that follows the libical API.
+- (TP) Brand new icalvcard library that follows the libical API.
 - draft-ietf-calext-eventpub-extensions-19 (RFC 9073) support added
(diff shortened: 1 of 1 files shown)


winterz:
I'll also add this same disclaimer to the Release Note on the github release page (whenever we finally release).

@ksmurchison @rsto let me know what you think.

winterz:
I just had a thought:  what if we do not build libicalvcard by default, but instead require a cmake option.

ie. `cmake -DLIBICAL_BUILD_LIBICALVCARD_TP=ON`

winterz:
> I just had a thought: what if we do not build libicalvcard by default, but instead require a cmake option.
> 
> ie. `cmake -DLIBICAL_BUILD_LIBICALVCARD_TP=ON`

I started implementing this and got stuck in libical-glib.  the vcard methods are integrated pretty tightly in libical-glib and it would be non-trivial to figure out how to conditionally build the glib bindings for libicalvcard.  well, maybe @mcrha could figure out how.  but that's not something I really want to put my time into.

mcrha:
Untested, but I guess it should be doable when you remove (do not add)
```cmake
# vCard files
list(
  APPEND API_FILES
  ...
```

in the `src/libical-glib/CMakeLists.txt` file, together with the other "vcard" references in there, when the vCard build is diasbled.

winterz:
> > I just had a thought: what if we do not build libicalvcard by default, but instead require a cmake option.
> > ie. `cmake -DLIBICAL_BUILD_LIBICALVCARD_TP=ON`
> 
> I started implementing this and got stuck in libical-glib. the vcard methods are integrated pretty tightly in libical-glib and it would be non-trivial to figure out how to conditionally build the glib bindings for libicalvcard. well, maybe @mcrha could figure out how. but that's not something I really want to put my time into.

if anyone is interested in taking it from here I put my WIP at #1301

winterz:
I'm going to merge this now and then create the 4.0 tag.
we can continue discussing our libicalvcard plans but I am getting anxious to get 4.0 out the door.

Issue #1311, opened 2026-04-28
Stack overflow in libical vCard parsing on deeply nested VCARD input


fuyu0425:
Hello libical maintainers,

I am reporting a reproducible failure: stack overflow in libical's vCard parsing path on a deeply nested VCARD input. I reproduced it by replaying the attached testcase `input.txt` against a locally built OSS-Fuzz address-sanitized target `libicalvcard_fuzzer` for project `libical`.

I am sending this as a public issue because the upstream repository does not currently publish a `SECURITY.md`, the README directs bug reports to the issue tracker, and the validated evidence is limited to a reproduced OSS-Fuzz ASAN crash. I have not separately validated release-build behavior or broader security impact.

Tested scope

- Reproduced on a locally built OSS-Fuzz target for project `libical`.
- Target name: `libicalvcard_fuzzer`.
- Upstream repository associated with the OSS-Fuzz project: `https://github.com/libical/libical`.
- This draft represents one grouped root-cause bucket built from `1` signature-deduped replay signature covering `4` original source POV entries.
- Released versions were not separately validated.

Observed result

- Reproduced against OSS-Fuzz target `libicalvcard_fuzzer` with sanitizer `address`.
- Observed crash: `SUMMARY: AddressSanitizer: stack-overflow /src/libical/src/libicalvcard/vcardderivedproperty.c:2785:38 in vcardproperty_string_to_kind`.
- The trace repeatedly recurses through `_parse_vcard` in `src/libical/src/libicalvcard/vcardparser.c:880`.
- Replay exit code: `1`.

Likely trigger / root cause

- The attached testcase is a deeply nested sequence of `BEGIN:VCARD` blocks.
- The replay trace suggests the parser recursively descends through nested vCards until the process stack is exhausted.
- The deepest repeated application frames stay in `_parse_vcard`, while the ASAN summary points at `vcardproperty_string_to_kind`. I have not fully reduced the testcase or proven the exact minimal root cause beyond this recursion pattern.

Current validated impact

- This report currently establishes only the reproduced behavior in the OSS-Fuzz address-sanitized target.
- I have not separately validated the same behavior in release builds or non-fuzz entrypoints.
- I have not established memory corruption or broader security impact beyond process crash / stack exhaustion.

Exact repro

1. Clone `google/oss-fuzz`.
2. Run the attached helper script with the OSS-Fuzz checkout and the attached testcase:

```bash
./repro-command.sh input.txt
```

[...]

PR #1312, opened 2026-04-28
libicalvcard: do not parse nested VCARD components

rsto:
This removes support for parsing nested VCARD components. There is no need for supporting that, VCARD components are not specified to be nested. The parser still supports parsing sequential VCARD components.

Related, this removes support for x-components in libicalvcard. There are no x-components defined in the RFCs and supporting them would require us to support nested components.

@winterz This should fix the issue reported in https://github.com/libical/libical/issues/1311. I successfully tested that the parser now rejects the input data of the issue report.

Please follow up with @ksmurchison for any changes to this PR, I might not find time to respond timely in the next days.


rsto:
glib build failed because it expected X-components still being supported. Now rerunning.

ksmurchison (approved):
This all looks correct to me

PR #1318, opened 2026-05-10
Increase the RRULE_SEARCH limit to 500

### LINKED ISSUE #1317: Wrong `is_daylight` returned by `icaltimezone_get_utc_offset` for data with very old `dtstart`
Consider the following test case, whose data is taken from the gnome-calendar test suite:
```
#include <stdlib.h>
#include <libical/ical.h>

#define VCALENDAR_DST \
  "BEGIN:VCALENDAR\n"                                  \
  "VERSION:2.0\n"                                      \
  "METHOD:PUBLISH\n"                                   \
  "BEGIN:VTIMEZONE\n"                                  \
  "TZID:StubTz\n"                                      \
  "BEGIN:STANDARD\n"                                   \
  "DTSTART:16011028T030000\n"                          \
  "RRULE:FREQ=YEARLY;BYDAY=-1SU;BYMONTH=10\n"          \
  "TZOFFSETFROM:+0200\n"                               \
  "TZOFFSETTO:+0100\n"                                 \
  "END:STANDARD\n"                                     \
  "BEGIN:DAYLIGHT\n"                                   \
  "DTSTART:16010325T020000\n"                          \
  "RRULE:FREQ=YEARLY;BYDAY=-1SU;BYMONTH=3\n"           \
  "TZOFFSETFROM:+0100\n"                               \
  "TZOFFSETTO:+0200\n"                                 \
  "END:DAYLIGHT\n"                                     \
  "END:VTIMEZONE\n"                                    \
  "BEGIN:VEVENT\n"                                     \
  "DTSTART;TZID=StubTz:20240603T103000\n"              \
  "DTEND;TZID=StubTz:20240603T114500\n"                \
  "SUMMARY:Test DST\n"                                 \
  "UID:test-dst@example\n"                             \
  "DTSTAMP:20240608T184316Z\n"                         \
  "END:VEVENT\n"                                       \
  "END:VCALENDAR\n"

int main () {
  int is_daylight = 0;
  putenv ("TZ=UTC");
  icalcomponent* comp = icalcomponent_new_from_string (VCALENDAR_DST);
  struct icaltimetype dt_start = icalcomponent_get_dtstart (comp);
  icaltimezone* zone = (icaltimezone*)icaltime_get_timezone (dt_start);
  icaltimezone_get_utc_offset (zone, &dt_start, &is_daylight);
  printf ("is_daylight: %d\n", is_daylight);
  return 0;
}
```
libical 4.0 gives
```
is_daylight: 0
```
which is incorrect.

Bisected to https://github.com/libical/libical/commit/d2060a6c94f7e936968de5bbfe4f81a9501aa7c9

winterz:
To fix the isdaylight test failure in the gnome-calendar test suite.

Fixes: #1317

### DIFF
--- a/docs/InternalLimits.md
+++ b/docs/InternalLimits.md
@@ -54,4 +54,4 @@ The internal limits and their default values are shown here:
 * Maximum number of times to search vtimezone rrules for an occurrence before the specified end year
 
       Kind = ICAL_LIMIT_RRULE_SEARCH
-      Default = 100 (one hundred)
+      Default = 500 (five hundred)
--- a/src/libical/icallimits.c
+++ b/src/libical/icallimits.c
@@ -24,7 +24,7 @@ static ICAL_GLOBAL_VAR size_t _MAX_VALUE_CHARS = 10485760;
 static ICAL_GLOBAL_VAR size_t _MAX_PROPERTY_VALUES = 500;
 static ICAL_GLOBAL_VAR size_t _MAX_RECURRENCE_SEARCH = 100000;
 static ICAL_GLOBAL_VAR size_t _MAX_RECURRENCE_TIME_STANDING_STILL = 50;
-static ICAL_GLOBAL_VAR size_t _MAX_RRULE_SEARCH = 100;
+static ICAL_GLOBAL_VAR size_t _MAX_RRULE_SEARCH = 500;
 
 size_t icallimit_get(icallimits_kind kind)
 {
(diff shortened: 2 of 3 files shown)


ksmurchison:
I guess we could/should allow expansion back to 1582 (start of Gregorian calendar), although that pre-dates time zones and probably would yield incorrect results anyways.  In fact, a time zone that originates in 1601 is nonsense and simply an academic exercise.  IIRC, the first known time zone emerged just after the US Civil War.

winterz:
my opinion is to keep things as they are.
but I would defer to your opinion if you disagree.  and send a patch :)

ksmurchison:
I'm fine with the current patch. Just stating that the bug itself is is real but unrealistic

PR #1328, opened 2026-05-20
Use global enums rather than strings to identify structure types

### LINKED ISSUE #1326: [4.0] warning: ‘__builtin_strncpy’ specified bound 5 equals destination size [-Wstringop-truncation]
These are the new claims from gcc I see here when compiling the 4.0 branch at commit c1b48b2263d7c1ff796c7fd9f9aee68aae23c58d:

```
..../libical/src/libical/icalvalue.c: In function ‘icalvalue_clone’:
..../libical/src/libical/icalvalue.c:76:5: warning: ‘__builtin_strncpy’ specified bound 5 equals destination size [-Wstringop-truncation]
   76 |     strncpy(clone->id, old->id, LIBICAL_ICALVALUE_ID_LENGTH);
      |     ^
..../libical/src/libicalvcard/vcardvalue.c: In function ‘vcardvalue_clone’:
..../libical/src/libicalvcard/vcardvalue.c:73:5: warning: ‘__builtin_strncpy’ specified bound 5 equals destination size [-Wstringop-truncation]
   73 |     strncpy(clone->id, old->id, LIBICAL_VCARDVALUE_ID_LENGTH);
```

The set `v->id[0] = 'X';` also seemed suspicious, but then I noticed it's in the `icalvalue_free()`, thus no big deal it's not `\0` terminated. It only confused me on the first look.

Personally, these "id" strings in the component, param, .... seem like a waste of space, considering structure size alignment can step in. Why not to use a single char, or some bool for "is/is-not valid", or even some global private enum telling what the structure is for, which I guess is the reason why these `id`-s exist in the implementation structures? You can save the `strcmp()` if it's an integer/enum too. Say something like this:

```c
typedef enum {
    PRIV_STRUCTURE_KIND_UNKNOWN,
    PRIV_STRUCTURE_KIND_VALUE,
    PRIV_STRUCTURE_KIND_FREED_VALUE,
    PRIV_STRUCTURE_KIND_PARAM,
    PRIV_STRUCTURE_KIND_FREED_PARAM,
    PRIV_STRUCTURE_KIND_PROP,
    PRIV_STRUCTURE_KIND_FREED_PROP,
    PRIV_STRUCTURE_KIND_COMPONENT,
    PRIV_STRUCTURE_KIND_FREED_COMPONENT
} PrivStructureKind;
```

winterz:
also saves some memory

fixes: #1326

### DIFF
--- a/src/libical/CMakeLists.txt
+++ b/src/libical/CMakeLists.txt
@@ -249,6 +249,7 @@ set(
   icalduration.c
   icalperiod.h
   icalperiod.c
+  icaltypes_p.h
   icaltypes.c
   icaltypes.h
   icalvalue.c
--- a/src/libical/icalcomponent.c
+++ b/src/libical/icalcomponent.c
@@ -25,13 +25,14 @@
 #include "icalrestriction.h"
 #include "icaltime_p.h"
 #include "icaltimezone.h"
+#include "icaltypes_p.h"
 
 #include <assert.h>
 #include <stdlib.h>
 #include <limits.h>
 
 struct icalcomponent_impl {
-    char id[5];
+    icalstructuretype id;
     icalcomponent_kind kind;
     char *x_name; /* also used for ICAL_IANA_COMPONENT */
     icalpvl_list properties;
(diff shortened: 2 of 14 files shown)


mcrha:
Looks good to me.

Are those impl structures in public API? If so, the changes in them would worth a soname version bump, because the size changed, thus anything using the libical should be rebuilt.

winterz:
There are no exported symbols in the libical/*impl.h nor are those files installed.

all they do is define some internal structures.

so I think we are safe.  agreed?

mcrha:
yup, in that case it's perfectly safe

PR #1333, opened 2026-05-25
icalvalue: avoid abs(INT_MIN) UB in utcoffset stringifier

Alearner12:
### Bug

`src/libical/icalvalue.c:941` uses `abs(data) == data` to determine the sign of a UTC offset:

```c
data = icalvalue_get_utcoffset(value);
if (abs(data) == data) {
    sign = '+';
} else {
    sign = '-';
}
```

`abs(INT_MIN)` is undefined behavior (C99 7.20.6.1). Furthermore, on two's-complement systems, `abs(INT_MIN)` typically returns `INT_MIN`, making the comparison `true` and incorrectly assigning a `+` sign to a negative offset.

### Reproduction

Built with `gcc -fsanitize=undefined`:

```
runtime error: negation of -2147483648 cannot be represented in type 'int'
data=-2147483648 -> sign=+  (Wrong: should be negative)
```

### Fix

Replace `abs(data) == data` with `data >= 0`. This is well-defined for all `int` values and correctly handles `INT_MIN`.

### Verification

Verified with the patched PoC:

```
data=-2147483648 -> sign=-  (Correct)
```

No sanitizer errors triggered. other path values (positive/negative offsets) remain unchanged.

### DIFF
--- a/src/libical/icalvalue.c
+++ b/src/libical/icalvalue.c
@@ -936,7 +936,7 @@ static char *icalvalue_utcoffset_as_ical_string_r(const icalvalue *value)
     str = (char *)icalmemory_new_buffer(9);
     data = icalvalue_get_utcoffset(value);
 
-    if (abs(data) == data) {
+    if (data >= 0) {
         sign = '+';
     } else {
         sign = '-';


Alearner12:
Sorry for the noise. On reflection  the parser at `icalvalue.c:565-580` can't produce `INT_MIN` from any input string, so the UB is only reachable through direct API misuse, not malformed iCalendar input. Closing.

winterz:
on the other hand, I don't like seeing UBSAN issues.

so I think we should fix as you suggested.  can you make a patch?
not only that but `data >=0` is a lot more obvious as to the intent of the code.

winterz:
nevermind.  I re-read your comment about "..API misuse, not malformed iCalendar input.." and I agree to close this one.

PR #1338, opened 2026-06-10
(ical|vcard)property.c: do not combine line fold and CRLF line ending

rsto:
This updates the line folding function for property content lines.

Before, a fold could get combined with the CRLF line ending, either by following a fold with the line ending (e.g. CR LF Space CR LF), or by interleaving the fold in between the ending (CR CR LF Space LF). The former is introducing an unnecessary fold and the latter is likely to break parsers.

Now, the content line is folded without the line ending and the final CRLF sequence is appended after folding. This avoids both pathological cases outlined above.

Thanks to @wolfsage for reporting this and providing a test case.


ksmurchison (approved):

winterz:
do you want this in the next 4.0.3 release?  if so, please move to the 4.0 branch

rsto:
@winterz done

winterz:
looks good.
please add a 1-line blurb into the CHANGELOG.md (under the [4.0.3] section) describing this change.
then feel free to merge

winterz:
forgot to mention that I plan to release version 4.0.3 this coming weekend.
let me know if there's any reason why I should delay that release.

rsto:
> looks good. please add a 1-line blurb into the CHANGELOG.md (under the [4.0.3] section) describing this change. then feel free to merge

Done! 

> forgot to mention that I plan to release version 4.0.3 this coming weekend.
> let me know if there's any reason why I should delay that release.

Nothing from my side. Thanks!

PR #1349, opened 2026-07-25
build: fix logic for processing deprecated options

jpalus:
new variable should be set to old variable's value not to old variable
name


winterz:
thank you

PR #1351, opened 2026-07-26
icaltime: fix 32-bit time_t upper bound off by two days (3.0 backport)

edmundlod:
Backport of #1350 to the `3.0` branch, which is what several
distributions — notably Debian's `libical3` package — still ship.
Same problem, same fix; see #1350 for the full writeup (root cause,
history, and testing). Summary below.

## Problem

On 32-bit `time_t` builds, `make_time()` rejects any date after
**2038-01-17**, two days before the real limit of a signed 32-bit
`time_t` — **2038-01-19T03:14:07 UTC** (`INT32_MAX` seconds since the
epoch). Rejected values fall through `icaltime_timegm()` as a bare
`return 0`, so `icaltime_as_timet_with_zone()` silently returns
1970-01-01 instead of the correct time_t value *or* an explicit error.

Any caller using `INT_MAX` as an "unbounded end" sentinel (a common
calendar idiom) gets that sentinel silently collapsed to epoch.

## Real-world impact

Found via Debian bug [#1142273](https://bugs.debian.org/1142273)
(cyrus-imapd FTBFS on i386). cyrus's `icalrecurrenceset_get_utc_timespan()`
passes `INT_MAX` as the outer search bound to
`icalcomponent_foreach_recurrence()`; on i386 that collapses to epoch,
so the search window shrinks to nothing-after-1970 and every
occurrence gets silently excluded — verified with a plain
`FREQ=WEEKLY;COUNT=3` rule from 2016, nowhere near the 2038 boundary.

## Fix

Exact boundary check (year/month/day/hour/min/sec) instead of the
padded day-level approximation, no arbitrary margin needed since
`icaltime_as_timet_with_zone()` already normalizes to UTC before this
check runs. Updates the two regression tests that had the old,
incorrect cutoff baked in as expected behavior.

## Testing

Same as #1350: round-trip test confirms `INT_MAX` now correctly
returns `INT_MAX` instead of `0` on an i386 build; full regression
suite passes; cyrus-imapd's real, previously-failing cunit test now
passes unmodified against a patched i386 `libical.so`.

---

*Disclosure: I used Claude (Anthropic) to help find and trace the root
cause on the 4.0 branch (#1350); this is a straightforward backport of
that fix to 3.0, done and tested the same way.*

### DIFF
--- a/src/libical/icaltime.c
+++ b/src/libical/icaltime.c
@@ -105,13 +105,17 @@ static int make_time(const struct tm *tm, int tzm, time_t *out_time_t)
     if (tm->tm_year > 138)
         return 0;
 
-    /* check for upper bound of Jan 17, 2038 (to avoid possibility of 32-bit arithmetic overflow) */
-    if (tm->tm_year == 138) {
-        if (tm->tm_mon > 0) {
-            return 0;
-        } else if (tm->tm_mday > 17) {
-            return 0;
-        }
+    /* 32-bit time_t cannot represent any instant after
+       2038-01-19T03:14:07 UTC (INT32_MAX seconds since the epoch). */
+    if (tm->tm_year == 138 &&
+        (tm->tm_mon > 0 ||
+         tm->tm_mday > 19 ||
+         (tm->tm_mday == 19 &&
+          (tm->tm_hour > 3 ||
+           (tm->tm_hour == 3 &&
+            (tm->tm_min > 14 ||
+             (tm->tm_min == 14 && tm->tm_sec > 7))))))) {
+        return 0;
     }
 #else
     /* We don't support years >= 10000, because the function has not been tested at this range. */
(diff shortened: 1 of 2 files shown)


winterz:
did you try cmake with the `-DLIBICAL_ENABLE_64BIT_ICALTIME_T=ON` option?
that is, if the target platform has __time64_t

edmundlod:
Debian decided that for its `i386` platform it would _not_ ABI transition to 64-bit time_t as policy. See [here](https://wiki.debian.org/ReleaseGoals/64bit-time#Decision). So that is not an option I'm afraid.


By the way, looking at the CMake files, I am wondering if the `-DLIBICAL_ENABLE_64BIT_ICALTIME_T=ON` works at all at the moment, since `CMakeLists.txt:485` validates the option using `SIZEOF___TIME64_T`, but that variable is only set by   `check_type_size()` inside `ConfigureChecks.cmake`, which is only included on line 551? Perhaps I am missing something, and it veers from the issue at hand.

winterz:
and 3.0 doesn't have the -DLIBICAL_ENABLE_64BIT_ICALTIME_T=ON option anyway.
and we still need to support 32-bit time properly.

I won't make a new 3.0 release but it does make sense to apply the patch there anyway so others can find it.

edmundlod:
Thank you, that is much appreciated!

PR #1369, opened 2026-09-06
Fix "now" linker support check.

jrmarino:
The current check for the "now" linker directive seems to be broken due to a copy-and-paste error.  Once it knows that "now" is supported, it adds "relro" to the linker flags.

These patches replace "relro" with "now", which is likely what was intended.


winterz:
indeed.  I wonder how I did that.

but, iirc I wanted partial relro.  So  should be:

```
check_linker_flag(C "-Wl,-z,relro" C_LINKER_SUPPORTS_RELRO)
if(C_LINKER_SUPPORTS_RELRO)
  set(CMAKE_EXE_LINKER_FLAGS "${CMAKE_EXE_LINKER_FLAGS} -Wl,-z,relro")
endif()
```

or even full relo:
```
check_linker_flag(C "-Wl,-z,relro,-z,now" C_LINKER_SUPPORTS_FULLRELRO)
if(C_LINKER_SUPPORTS_FULLRELRO)
  set(CMAKE_EXE_LINKER_FLAGS "${CMAKE_EXE_LINKER_FLAGS} -Wl,-z,relro,-z,now")
endif()
```

for C++ too, of course.

thoughts? could you test if either of those work for you.

jrmarino:
I found this on omnios.  The illumos linker supports "now" but not "relro".  So libical failed the "relro" check (as it should have) but passed the "now" check, then broke because "relro" is not supported.  

So the first example you posted already exists.

The second example you posted should work on omnios.

winterz:
ignore me.  you're correct about the typo.

winterz:
thanks very much.

PR #1373, opened 2026-09-07
Avoid `icalstrarray_clone()` allocating its capacity incrementally

minichma:
Some performance improvement of icalstrarray_clone(): Allocate the whole array capacity right at the beginning rather than starting with a single chunk and resizing it incrementally as elements are added as done previously. The old behavior resulted in O(N^2) complexity, which made a difference in case of large arrays, as seen in some fuzz tests.

### DIFF
--- a/src/libical/icalstrarray.c
+++ b/src/libical/icalstrarray.c
@@ -148,11 +148,29 @@ icalstrarray *icalstrarray_clone(icalstrarray *array)
         return NULL;
     }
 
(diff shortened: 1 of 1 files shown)


winterz (approved):
I tested locally and everything looks great.
we'll see what ossfuzz says in the upcoming days.

minichma:
> we'll see what ossfuzz says in the upcoming days.

This is just some minor improvement, it doesn't fix oss issue though, as it doesn't reduce the allocation level. Might just bring some performance benefit.

PR #1381, opened 2026-09-09
Remove ancient workaround for malformed `TZID` param

### LINKED ISSUE #1376: Parser: Special handling of `:` in `TZID` parameter broken
libical implements [special handling for `TZID` parameters](https://github.com/libical/libical/blob/ad6b6bd6175e1e34af569d37136edf6bd77cdb67/src/libical/icalparser.c#L956-L1017) in order to support certain non-conformant data (the following code comment is dating back to 2012):

```
For example, Google Calendar will send back stuff like this:
DTSTART;TZID=GMT+05:30:20120904T020000
```

In the current form, the special handling prevents the parser from properly dealing with certain input. Consider the following case, which was derived from some fuzz test.

```
RESOURCES;TZID=x:A;TZID=:B;TZID=:abc
```

The syntactically correct interpretation would be as follows:
```
parameters:
  TZID   = x
value:
  A;TZID=:B;TZID=:abc
```

Instead, it is interpreted the same as the following.

```
RESOURCES;TZID="x:A";TZID="x:A;TZID=:B";TZID="":abc
```

i.e.

```
parameters:
  TZID   = x:A
  TZID   = x:A;TZID=:B
  TZID  = <empty>
value:
  abc
```

Particularly note the `x:A` appearing multiple times. Actually the problem can cause O(N^2) memory consumption, as additional `;TZID=:x` spans would each cause both, an additional parameter being allocated as well as the other parameters' values being longer.

The problem popped up because some fuzz tests failed with out of memory exceptions.

Is the workaround for the broken input still needed? If not, the simples fix would probably be to just remove it.

minichma:
Remove a workaround dating back to 2012 that addressed malformed `TZID` params like the following (note the unescaped colon in `05:30`):

```
DTSTART;TZID=GMT+05:30:20120904T020000
```

The workaround itself was faulty, causing O(N^2) memory consuption in certain cases as seen in some fuzz tests and discussed in #1376. This PR removes it without replacement, assuming it isn't needed anymore and given that it was conflicting wit the RFC.

closes #1376


winterz (approved):
I tested this patch in KDE:  the kcalendarcore library tests all passed.  then I ran KOrganizer with the patch and saw no problems.

wonderful.  thank you.

PR #1388, opened 2026-09-10
Fix libicalvcard out-of-memory fuzz issue

winterz:
see https://oss-fuzz.com/testcase-detail/6073213411721216


winterz:
@minichma any chance you can look at the associated ossfuzz issue and measure if a patch like this one is any help?
I mean measure the memory consumption.

ksmurchison (approved):

minichma:
> @minichma any chance you can look at the associated ossfuzz issue and measure if a patch like this one is any help? I mean measure the memory consumption.

Sure. Maximum allocation level:

* Previous (4edd39a34a7b67ca6ab68667c84ac19ae890dd43): 1.450.283.384 Bytes
* New (230bad51a86d0e096a7f97bac485388d6ab18fad): 22.186 Bytes 👍 

See https://github.com/libical/libical/tree/work/minichma/fuzz_print_mem (this doesn't build on all platforms, so jfyi, but nothing to merge)

winterz:
great. looks like this patch will help a great deal.

on second thought:  I don't know that we need an internal limitation getter/setter on the length or a property name (for libicalv5). does anyone else?  I think hard-coding at 100 chars is plenty.

IOW: should I remove the comment "//TODO: 5.0 icallimit_get(ICAL_LIMIT_PROPERTY_NAME_LENGTH);"

ksmurchison:
I think 100 char property name is plenty

winterz:
> I think 100 char property name is plenty

thanks.  I'll change the comment to "we think 100 chars for a property name is plenty" (or some-such)

PR #1404, opened 2026-09-14
Fix out-of-memory ossfuzz issue 560953111

winterz:
Limit the number of ical parameter values to ICAL_LIMIT_PROPERTY_VALUES.
For this test, the memory allocation is reduced from ~2.5Gb to 3.5Mb.

see https://oss-fuzz.com/testcase-detail/4868613300092928


winterz:
@minichma can you do the before / after memory allocation measurement test for this one?

PS. I could see some things we could do differently in the parser that would be another way to solve this but when I do that I break other tests.  in some ways we are being too lenient.  but I don't want to deal with that (at least not until libical5)

minichma:
@winterz Sure. The level of memory allocated via icalmemory_new_buffer() et al is the same in both cases:

new (262158bad2b1394ea0d4de18d97c3ad33fe226cc): 249641195
old (11ba99774bd59d0e715ab0271043286b9916ed3e): 249641195

But when looking at the stack trace on oss-fuzz it seems, the failing allocation is not handled via icalmemory_new_buffer(), so the 'test-malloc' measurement won't work.

```
  | #0 0x57df6a9d5814 in __interceptor_malloc /src/llvm-project/compiler-rt/lib/asan/asan_malloc_linux.cpp:67:3
  | #1 0x57df6aacf273 in operator new(unsigned long) cxa_noexception.cpp:0
  | #2 0x57df6a8d3632 in main /src/llvm-project/compiler-rt/lib/fuzzer/FuzzerMain.cpp:20:10
  | #3 0x7cf2473bb082 in __libc_start_main /build/glibc-B3wQXB/glibc-2.31/csu/libc-start.c:308:16
  |  
  | SUMMARY: libFuzzer: out-of-memory
  |  
```

winterz:
thanks.
this one is proving to be difficult to solve for me.

I'll try again another day . unless someone else wants to give it a go.

minichma:
I tried to understand the crash output (with the help of ChatGPT). To my understanding the amount of memory allocated by libical is not the problem but the amount allocated by the fuzzer/asan is:

> Live Heap Allocations: 25058341 bytes in 26401 chunks; quarantined: 168284030 bytes in 9193076 chunks; 14380370 other chunks; total chunks: 23599847; showing top 95% (at most 8 unique contexts)

The quarantined memory seems to be the one that has already been freed but is still tracked by asan. Not sure, this understanding is correct though.

Anyhow, the number of allocations is high by design. The `rESoURCES` property in line 4 of the test case has one parameter with a significant number of values (41.970; the number doesn't seem to be limited). The property itself is multivalued and has a significant number of values as well (~1k but a limit of 500 is applied by libical). In contrast to the icalparameter type, the icalproperty type in libical doesn't natively support multiple values, so the workaround is to allocate each value as a separate property. Each copy also has a copy of all the parameters, resulting in 500*41970 = 20.985.000 values being allocated. Per value we have >= 1 memory allocation, as we not only allocate the value itself but also need to grow the array it is referenced by.

As the asan seems to track all the allocations, this seems to result in the overflow.

Potential improvements:
* Build multi-value support into `icalproperty` - probably a breaking change, so nothing for v4
* Apply a limit to the max number of values per parameter, maybe around [here](https://github.com/libical/libical/blob/11ba99774bd59d0e715ab0271043286b9916ed3e/src/libical/icalderivedparameter.c.in#L239-L293)
* Avoid frequent reallocations of growing buffers, e.g. create arrays having the right capacity from the beginning rather than growing incrementally.
* Improve the sanity check and reject malformed input (good to have but low impact for the particular problem)

A related problem that might not impact this particular PR but the related #1388 originates from the way how temporary buffers are managed. `icalmemory` manages a ring buffer of 2500 entries. If a buffer is added to it, it is only released after 2500 other buffers have been added. The size of the 2500 adds up, so if many of them are large, the amount of allocated memory can be significant. In #1388 a big part of the content line could be added to the temp buffer. As this happened many times, the amount of memory being allocated could be roughly proportional to 2500 times the size of the input line. For that reason and also for the inevitable concurrency challenges that come with such a global ring buffer, I feel it would be good to get rid of the icalmemory ring buffer altogether, at least in v5. But as mentioned before, this topic doesn't have immediate impact on the test case of this PR.

winterz:
great analysis.  my patch #1408 verifies that we aren't using all that 2GB.  not even half.
even so,  those fuzz emails are irritating so we should try a little bit.

I've been concentrating on the possible solution "Improve the sanity check and reject malformed input (good to have but low impact for the particular problem)" and getting nowhere without breaking some tests -- this is what I meant by "we're being too lenient".  anyway, something for v5.

narrowing things down we have 2 options that might work in v4:
* Apply a limit to the max number of values per parameter,
* Avoid frequent reallocations of growing buffers

the first option I will try myself.
the second option would need to be something for you to try, if you want.  If not, that's ok too.

minichma:
> the first option I will try myself.
> the second option would need to be something for you to try, if you want. If not, that's ok too.

Sounds good. Will put the second one on my list. It's not the core problem though, so I guess it won't be a problem if it takes some time.

winterz:
this one is ready for another review.
sorry about the whitespace changes.

minichma (approved):

PR #1405, opened 2026-09-14
clang-tidy generator.c - clang-analyzer-security.ArrayBound (#1398)

### LINKED ISSUE #1398: generator.c - clang-analyzer-security.ArrayBound
Running clang-tidy v22.1.8 today I'm seeing this warning from generator.c

```
clang-analyzer-security.ArrayBound:
  + src/libical-glib/tools/generator.c:  line 1871:Out of bound access to memory after the end of the heap area
```

mcrha:
Improve real type parsing from libical-glib API .xml files.

Closes #1398


winterz:
```/home/winterz/projects/libical/libical/src/libical-glib/tools/generator.c:1883:24: warning: Use of memory allocated with size zero [unix.Malloc]```  reported by scan-build.

I think we just forget the whole thing and I'll tell clang-tidy to ignore the warning.  do you agree?  something in the glib functions must be confusing clang-tidy.

mcrha:
There might definitely something confuse it, because with the check `start <= end` (that `g_return_val_if_fail()` can be removed in compile time when glib is configured with special option, but it did not trigger here), then the `res` is always at least two bytes long, thus writing to byte `[1]` is correct, regardless how long the input string was - and it always gets at least one character from the input string, which is nonempty. My test code (in https://github.com/libical/libical/issues/1398#issuecomment-5666978276 ) tries even non-sense to pass in and nothing reproduces the problem the analyzer thinks can happen.

In any case, this change improves some parsing, even it's not really needed due to the controlled input strings, thus I would merge this and rather mute the `[unix.Malloc]` claim than the `ArrayBound` claim.

winterz:
ok I'll merge this and then follow-up with the change to silence the warning.

PR #1407, opened 2026-09-15
generator.c - silence the unix.Malloc false-positive from clang-analyzer

winterz:


winterz:
grr.. now cppcheck is reporting:

```
generator.c:1417,style,knownConditionTrueFalse,Condition 'trueType' is always false
generator.c:1847,style,unusedVariable,Unused variable: res  (obviously.  I'll fix that)
generator.c:2061,style,knownConditionTrueFalse,Condition '!trueType' is always true
```

winterz:
@mcrha could you look at the trueType issues from cppcheck?
let's make that a follow-on patch to this.

winterz:
> @mcrha could you look at the trueType issues from cppcheck? let's make that a follow-on patch to this.

nevermind.  I cannot reproduce any longer.

mcrha:
Oh, you silence the warnings by not compiling the code... I hoped in a suppression file or comment (like in case of the Coverity Scan, which is sensitive to certain comments).

I guess the cppcheck (in combination with clang-analyzer) was correct, because the function always returns `NULL` in that case, which breaks the checks for the returned value.

mcrha:
...which breaks the checks for the returned value in the callers of that function.

winterz:
> ...which breaks the checks for the returned value in the callers of that function.

I couldn't figure out any other way to silence the false-positive.  I tried various NOLINT comments without any luck.

would you like me to revert? 

this is a static checker tool so I didn't think it mattered about runtime.

mcrha:
No no, do not revert. My comment was more like "yay, this is surprising", but nothing else about it.

PR #1408, opened 2026-09-15
Add ability to set a max memory allocation limit with MEMORY_CONSISTENCY

winterz:
Add test-malloc API for setting/getting the maximum memory
allocation limit. The calling program exits with code=1 if
that limit is exceeded.

In the fuzzer test programs:
 - use the new test-malloc API to fail if more than 1GB of memory
   allocation is attempted.
 - print the amount of memory allocated by the test


winterz:
In my local testing, none of our current fuzz tests exceed 1GB memory allocation.

With this patch I can try various scenarios in future fuzz tests to see how the memory allocation changes.

src/test/libicalvcard/vcardcomponent_fuzz1.c
```diff
@@ -33,6 +37,12 @@ int main(int argc, const char *argv[])
     void *data = NULL;
     vcardcomponent *card;
 
+#if defined(MEMORY_CONSISTENCY)
+    size_t maxMem = TEST_MAX_MEMORY;
+    icalmemory_set_mem_alloc_funcs(&test_malloc, &test_realloc, &test_free);
```
minichma, line 42: icalmemory_set_mem_alloc_funcs shouldn't be needed, because test_malloc et al are the default when MEMORY_CONSISTENCY is set.

src/test/icalcomponent_fuzz1.c
```diff
@@ -33,6 +36,12 @@ int main(int argc, const char *argv[])
     void *data = NULL;
     icalcomponent *comp;
 
+#if defined(MEMORY_CONSISTENCY)
+    size_t maxMem = TEST_MAX_MEMORY;
+    icalmemory_set_mem_alloc_funcs(&test_malloc, &test_realloc, &test_free);
```
minichma, line 41: icalmemory_set_mem_alloc_funcs shouldn't be needed

src/test/icalcomponent_fuzz2.c
```diff
@@ -33,6 +36,12 @@ int main(int argc, char const *argv[])
     void *data = NULL;
     icalcomponent *comp1, *comp2;
 
+#if defined(MEMORY_CONSISTENCY)
+    size_t maxMem = TEST_MAX_MEMORY;
+    icalmemory_set_mem_alloc_funcs(&test_malloc, &test_realloc, &test_free);
```
minichma, line 41: icalmemory_set_mem_alloc_funcs shouldn't be needed

minichma (approved):

src/test/icalcomponent_fuzz2.c
winterz: I tried that.  when I did, the max setting didn't work at all and the print for the amount of memory was zero.

I fought quite a bit with it.  maybe you know what I forgot.

minichma:
> In my local testing, none of our current fuzz tests exceed 1GB memory allocation.
> 
> With this patch I can try various scenarios in future fuzz tests to see how the memory allocation changes.

I believe the out-of-memory is mainly caused by the large number of allocations. The oss-fuzz stacktrace states ASAN option `redzone=64`. This seems to cause an unused block of memory of 64 bytes around each allocation. Given the total number of chunks of 23599847, this would result in ~1.5G just for the redzone.

I've listed a few ideas for reducing the number of allocations in https://github.com/libical/libical/pull/1404#issuecomment-5669715490. The most effective measure would be adding multi-value support to icalproperty, but that would be a breaking change. Limiting the number of values per parameter (and probably others) would be more than helpful as well and would be easy.

src/test/icalcomponent_fuzz2.c
minichma: Oh, I thought it wouldn't be needed, because it worked on my machine. But that's probably because I'm compiling with STATIC_ONLY. I remember facing a similar problem when initially working on that stuff quite some time ago. Maybe test-malloc.c is compiled into both, the lib and the executable. If not set from the executable, the version from the lib is in effect. But when calling testmalloc_set_max_memory et al from the executable, you'd be setting the backing field from the executable version, which wouldn't be picked up by the version from the lib. Setting the functions explicitly is an effective workaround. Sorry for the confusion.

winterz:
> > In my local testing, none of our current fuzz tests exceed 1GB memory allocation.
> > With this patch I can try various scenarios in future fuzz tests to see how the memory allocation changes.
> 
> I believe the out-of-memory is mainly caused by the large number of allocations. The oss-fuzz stacktrace states ASAN option `redzone=64`. This seems to cause an unused block of memory of 64 bytes around each allocation. Given the total number of chunks of 23599847, this would result in ~1.5G just for the redzone.
> 
you finally got through to me.  yes, now I understand and I think you are absolutely correct about the number of allocations.

test-malloc allows us to limit the number of allocations.  one day we could enforce that in the fuzzer tests just like I'm trying to do for the total memory allocation limit.  thank you for test-malloc btw.  I never realized how useful it is before now :)

minichma:
Thanks for the kudos :-) Btw, test-malloc adds some overhead per allocation as well, on my machine 16B, but those are not considered in `global_testmalloc_statistics.mem_allocated_*`.

PR #1410, opened 2026-09-17
Fixes for undefined sanitizer issues

winterz:


winterz:
a couple things with this patch to lookout for:

- test-malloc.  please double check the couple changes there.  mostly I told ubsan to skip the code in test-malloc because I couldn't understand how to fix.  @minichma I can work with you if interested.

- the new makeMask() static function exists to isolate a ubsan issue about left-shifting by too many bits.  if someone knows better how to make a bitmask please let me know

winterz:
in particular, this patch should fix ossfuzz issue https://oss-fuzz.com/testcase-detail/4549369773490176 (an integer-overflow in icalvalue_new_from_string)

src/test/test-malloc.c
```diff
@@ -167,7 +180,11 @@ void test_free(void *p)
 
     free(hdr);
 
-    global_testmalloc_statistics.mem_allocated_current -= old_size;
+    if (global_testmalloc_statistics.mem_allocated_current > old_size) {
+        global_testmalloc_statistics.mem_allocated_current -= old_size;
```
minichma: Just a comment, not a problem: If we'd have ran into an overflow in a previous call to malloc, then this subtraction would result in an incorrect result. Its a conflict, either we accept the overflow when adding, or we risk to have incorrect values when subtracting. I feel, I'd rather go for accepting the overflow, because its less complex and IMHO good enough, as we're most often talking about an uint64. Even if there *is* an overflow in the 32-bit case, probably nothing bad would happen. Maybe maintaining an additional flag that's indicating that there was an overflow could be useful, but even that probably wouldn't be worth the effort.

minichma (approved):
I don't know much about the sanitizer, so I can't comment on the attributes, etc., just on the 'actual code' changes.

src/libical/icalvalue.c
minichma: Doesn't `icalvalue_new_utcoffset()` take an `int`? Could this cause new sanitizer issues?

minichma:
> @minichma I can work with you if interested.

Sure, would be happy to. What should I do?

src/test/test-malloc.c
winterz: > Just a comment, not a problem: If we'd have ran into an overflow in a previous call to malloc, then this subtraction would result in an incorrect result. Its a conflict, either we accept the overflow when adding, or we risk to have incorrect values when subtracting. I feel, I'd rather go for accepting the overflow, because its less complex and IMHO good enough, as we're most often talking about an uint64. Even if there _is_ an overflow in the 32-bit case, probably nothing bad would happen. Maybe maintaining an additional flag that's indicating that there was an overflow could be useful, but even that probably wouldn't be worth the effort.

agree. I'll remove those changes in test-malloc

winterz:
> > @minichma I can work with you if interested.
> 
> Sure, would be happy to. What should I do?

I use clang on linux (Mac should work too).  for Windows I don't know.
In theory all you need to do is pass `-DLIBICAL_DEVMODE_UNDEFINED_SANITIZER=True` to cmake and compile/test normally.  you might also want  `-DLIBICAL_BUILD_TESTING_BIGFUZZ=True` for the fuzzy tests.

and put this in your environment first:

```
export UBSAN_OPTIONS=allocator_release_to_os_interval_ms=500:halt_on_error=1:handle_abort=2:handle_segv=2:handle_sigbus=2:handle_sigfpe=2:handle_sigill=2:print_stacktrace=1:print_summary=1:print_suppressions=0:silence_unsigned_overflow=1:symbolize=1:use_sigaltstack=1
```

src/test/test-malloc.c
winterz: done in updated patch

src/libical/icalvalue.c
winterz: good catch.
should be fixed in updated patch

Issue #504, opened 2021-08-15
Generate source using flex+bison


winterz:
this may be impossible in libicalvcal.
should be possible in libicalss

I started in the flexbison branch if anyone wants to pick it up.
So far only in libcalss it compiles and links; however the regression test core dumps

PR #1416, opened 2026-09-18
Fix memory statistics handling in regression-utils to avoid size_t underflow

minichma:
regression-utils implements functionality to detect memory leaks of individual tests. The implementation was resetting the memory allocation level to 0 after each test, which caused the final free() at the end of the test run to cause underflows. Those were snoozed by compiler directives.

This PR fixes the mem statistics handling and removes the now obsolete compiler directives.


minichma:
@winterz I'm maintaining a ubuntu-based dockerfile for libical development and that might be useful also for others. Do we have a good place where I could share it? Should we include it in the code base?

winterz:
> @winterz I'm maintaining a ubuntu-based dockerfile for libical development and that might be useful also for others. Do we have a good place where I could share it? Should we include it in the code base?

yes, I think having Dockerfiles in the code base makes sense.
feel free to put it where it makes sense .. how about in top-level/docker/Dockerfile-ubuntu ?

winterz:
if I set:
```
UBSAN_OPTIONS=allocator_release_to_os_interval_ms=500:halt_on_error=1:handle_abort=2:handle_segv=2:handle_sigbus=2:handle_sigfpe=2:handle_sigill=2:print_stacktrace=1:print_summary=1:print_suppressions=0:silence_unsigned_overflow=1:symbolize=1:use_sigaltstack=1
```

then I get:
```
    Start 1: regression
1/1 Test #1: regression .......................***Failed    0.05 sec
/home/winterz/projects/libical/libical/src/test/test-malloc.c:55:27: runtime error: member access within null pointer of type 'struct testmalloc_hdrlayout'
    #0 0x00000045f6da in test_malloc /home/winterz/projects/libical/libical/src/test/test-malloc.c:55:27
    #1 0x7f92e4ed0c5b in icalmemory_new_buffer /home/winterz/projects/libical/libical/src/libical/icalmemory.c:324:9
    #2 0x7f92e4f10831 in icaltimezone_set_zone_directory /home/winterz/projects/libical/libical/src/libical/icaltimezone.c:2382:28
    #3 0x00000043ee0a in main /home/winterz/projects/libical/libical/src/test/regression.c:7851:5
    #4 0x7f92e480a680 in __libc_start_call_main (/lib64/libc.so.6+0x3680) (BuildId: 5bd941be836f538fe5e10eff508f7f5dd94905a6)
    #5 0x7f92e480a797 in __libc_start_main@GLIBC_2.2.5 (/lib64/libc.so.6+0x3797) (BuildId: 5bd941be836f538fe5e10eff508f7f5dd94905a6)
    #6 0x000000411374 in _start (/home/winterz/projects/libical/libical/build-test2ubsan-ubsan/src/test/regression+0x411374) (BuildId: 825e370e8ef11654bd45231f40190263074be083)
```

minichma:
> if I set:
> ...
> then I get:

Oh no, I'll check again

winterz:
> > if I set:
> > ...
> > then I get:
> 
> Oh no, I'll check again

also make sure to use clang.  gcc's ubsan is not nearly as good as clang's (in my experience)

winterz:
> > > if I set:
> > > ...
> > > then I get:
> > 
> > 
> > Oh no, I'll check again
> 
> also make sure to use clang. gcc's ubsan is not nearly as good as clang's (in my experience)

come to think of it, the buildsystem should fail if you try to use the undefined sanitizer with gcc.

minichma:
Could reproduce and fix the issue. ubsan still issues a runtime error, but this time in vcc.y. Didn't dig into that one.

```
vcc.y:553:16: runtime error: implicit conversion from type 'int' of value -1 (32-bit, signed) to type 'char' changedthe value to 255 (8-bit, unsigned)
```

winterz:
> Could reproduce and fix the issue. ubsan still issues a runtime error, but this time in vcc.y. Didn't dig into that one.
> 
> ```
> vcc.y:553:16: runtime error: implicit conversion from type 'int' of value -1 (32-bit, signed) to type 'char' changedthe value to 255 (8-bit, unsigned)
> ```

yeah, I forgot to mention that I ignored libicalvcal issues.

maybe you can try to fix that one?
or at least add the `__attribute__` to tell the compiler to ignore it.

winterz (approved):
wonderful

minichma:
> > Could reproduce and fix the issue. ubsan still issues a runtime error, but this time in vcc.y. Didn't dig into that one.
> > ```
> > vcc.y:553:16: runtime error: implicit conversion from type 'int' of value -1 (32-bit, signed) to type 'char' changedthe value to 255 (8-bit, unsigned)
> > ```
> 
> yeah, I forgot to mention that I ignored libicalvcal issues.
> 
> maybe you can try to fix that one? or at least add the `__attribute__` to tell the compiler to ignore it.

I don't feel comfortable with that code and never worked with yacc, but it seems the fix was easy, so gave it a try. How do we usually deal with that yacc-generated code? I guess we don't use yacc a lot, do we? I just applied the change to both, vcc.c and vcc.y. Is this ok?

winterz:
> > > Could reproduce and fix the issue. ubsan still issues a runtime error, but this time in vcc.y. Didn't dig into that one.
> > > ```
> > > vcc.y:553:16: runtime error: implicit conversion from type 'int' of value -1 (32-bit, signed) to type 'char' changedthe value to 255 (8-bit, unsigned)
> > > ```
> > 
> > 
> > yeah, I forgot to mention that I ignored libicalvcal issues.
> > maybe you can try to fix that one? or at least add the `__attribute__` to tell the compiler to ignore it.
> 
> I don't feel comfortable with that code and never worked with yacc, but it seems the fix was easy, so gave it a try. How do we usually deal with that yacc-generated code? I guess we don't use yacc a lot, do we? I just applied the change to both, vcc.c and vcc.y. Is this ok?

yes, that's the correct solution. 
 
at times I have attempted to use flex/bison at configure time to generate the vcc.c .  I haven't succeeded yet. see the work/winterz/flexbison branch and https://github.com/libical/libical/issues/504

winterz (approved):
thank you.

PR #1421, opened 2026-09-21
Fix ossfuzz issue cpu timeout in vcardtextlist_new_from_string()

winterz:
Limit the number of iterations for building the text list to
ICAL_LIMIT_PROPERTIES (currently at 10,000).

see https://oss-fuzz.com/testcase-detail/6003677710843904


winterz:
this patch reduces runtime of the ossfuzz test by about 20x.

but I have no idea if 100,000 chars is enough.
I could try 250,000 and 500,000.

I already timed 1,000,000 and its pretty slow indeed.

winterz:
Here's my timing tests:  I don't know why oss-fuzz is complaining. 

my old laptop says:

* 100000 char limit  => 1.790s;   max memory allocation level: 2785688
* 250000 char limit  => 9.512s;   max memory allocation level: 5185688
* 500000 char limit  => 15.349s;  max memory allocation level: 6582680
* 1000000 char limit => 15.138s; max memory allocation level: 6582680
 * unlimited (SIZE_MAX) =>14.787s; max memory allocation level: 6582680

I was wrong earlier when I said "20x faster with this patch".  more like 10x.

waiting for your feedback @rsto and @ksmurchison

